PR #1, opened 2021-06-17
initial commit

s-urbaniak:
/cc @stlaz


images/ci/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
```
slaskawi, line 1: How about using Go 1.16? I've seen it in other places...

slaskawi (approved):

images/ci/Dockerfile
s-urbaniak: oh good catch, indeed :+1:

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cert-manager-operator/pkg/controller/deployment"
```
stlaz, line 15: I like to take the local imports into their own separate block, it sometimes makes rebasing easier, especially on someone else's changes

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+	configClient, err := configv1client.NewForConfig(cc.KubeConfig)
+	if err != nil {
+		return err
+	}
+
+	genericOperatorClient, dynamicInformers, err := genericoperatorclient.NewClusterScopedOperatorClient(
+		cc.KubeConfig, operatorv1.GroupVersion.WithResource("certmanagers"))
+	if err != nil {
+		return err
+	}
```
stlaz, line 44: maybe we can have our own operator client created later but this is fine for now

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+
+	clusterOperator, err := configClient.ConfigV1().ClusterOperators().Get(ctx, "cert-manager-operator", metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	versionRecorder := status.NewVersionGetter()
```
stlaz, line 51: In the CAO I added a comment explaining why the `SetVersion()` is called here because it took me some digging to realize what's going on. Maybe you'd like to include it as well.

https://github.com/openshift/cluster-authentication-operator/blob/ca4d280f750ed8f04bbee47422e6e29da9064d0c/pkg/operator/starter.go#L154-L156

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
+	"github.com/openshift/library-go/pkg/operator/status"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	TargetNamespace   = "openshift-cert-manager"
+	OperatorNamespace = "openshift-cert-manager-operator"
```
stlaz, line 26: suggestion: move all constants into a single folder, like `api` or something similar to KAS-o's https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/operatorclient/interfaces.go

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+		kubeClient,
+		kubeInformersForTargetNamespace.InformersFor(TargetNamespace),
+		configClient.ConfigV1().ClusterOperators(),
+		cc.EventRecorder,
+		versionRecorder,
+	)
+
+	configInformers.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+	kubeInformersForTargetNamespace.Start(ctx.Done())
```
stlaz, line 88: suggestion:
```golang
for _, informer := interface{Start(<-chan struct{})}{
  configInformers,
  dynamicInformers,
  kubeInformersForTargetNamespace
} {
    informer.Start(ctx.Done())
}
```

pkg/operator/starter.go
```diff
@@ -0,0 +1,95 @@
...
+
+	go statusController.Run(ctx, 1)
+	go deploymentController.Run(ctx, 1)
+
+	<-ctx.Done()
+	return fmt.Errorf("stopped")
```
stlaz, line 94: won't this cause the operator to exit with an error?

pkg/tools/tools.go
```diff
@@ -0,0 +1,8 @@
+// +build tools
```
stlaz, line 1: I prefer `dependencymagnet`, it's more self-explanatory IMO

pkg/version/version.go
```diff
@@ -0,0 +1,47 @@
...
+		prometheus.GaugeOpts{
+			Name: "openshift_cert_manager_operator_build_info",
+			Help: "A metric with a constant '1' value labeled by major, minor, git commit & git version from which OpenShift Cert Manager Operator was built.",
+		},
+		[]string{"major", "minor", "gitCommit", "gitVersion"},
+	)
+	buildInfo.WithLabelValues(majorFromGit, minorFromGit, commitFromGit, versionFromGit).Set(1)
+
+	prometheus.MustRegister(buildInfo)
+}
```
stlaz, line 47: this might be getting reported here?

https://github.com/openshift/library-go/blob/28a7a4065a9e0070ef60c661f37903fa875ac332/pkg/controller/controllercmd/builder.go#L227-L241

stlaz (commented):
more or less just suggestions, looking good so far

pkg/operator/starter.go
s-urbaniak: yes agreed

pkg/operator/starter.go
s-urbaniak: sounds good, i think i copy-pasted this from somewhere

pkg/operator/starter.go
s-urbaniak: yea i saw that we are maintaining a whole struct in authentication-operator, sounds fine to me.

pkg/version/version.go
s-urbaniak: i think i took this over from somewhere, but you know what it just occured to me: we really should dependency-inject the registry and omit the global init.

s-urbaniak: and let me recheck if we hit the above library-go function

pkg/operator/starter.go
s-urbaniak: sure thing :+1:

pkg/operator/starter.go
s-urbaniak: added

pkg/tools/tools.go
s-urbaniak: definitely, changed :+1:

pkg/operator/starter.go
s-urbaniak: agreed, let's leave this out, i have this from https://github.com/openshift/cluster-kube-storage-version-migrator-operator/blob/8115240c5c0ceab6102682ffa5a0f9945c6151b9/pkg/operator/starter.go#L109 and was assuming RunOperator is not supposed to be finished directly. I guess we should it fix it there :thinking:

pkg/operator/starter.go
s-urbaniak: after thought: let's leave this out for now as the unstuctured.Unstructured informer will otherwise complain with `failed to list *unstructured.Unstructured: the server could not find the requested resource`.

pkg/version/version.go
s-urbaniak: ok, nice, yes, the builder auto-injects the build info already :+1: 
```
$ curl -s -k -H "Authorization: Bearer $(kubectl -n openshift-monitoring get secret $(kubectl -n openshift-monitoring get sa prometheus-k8s -o jsonpath='{.secrets[0].name}') -o jsonpath='{.data.token}' | base64 -d)" https://localhost:8443/metrics | grep build_info
# HELP openshift_cert_manager_operator_build_info [ALPHA] A metric with a constant '1' value labeled by major, minor, git version, git commit, git tree state, build date, Go version, and compiler from which openshift-cert-manager-operator was built, and platform on which it is running.
# TYPE openshift_cert_manager_operator_build_info gauge
openshift_cert_manager_operator_build_info{buildDate="2021-06-18T15:47:43Z",compiler="",gitCommit="49c5269",gitTreeState="",gitVersion="v0.0.0-unknown",goVersion="",major="",minor="",platform=""} 1
```

PR #8, opened 2021-06-24
apis/config: add config type

s-urbaniak:
/cc @stlaz


apis/config/v1alpha1/types_certmanager.go
```diff
@@ -0,0 +1,34 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Certmanager specifies cluster-wide settings for cert-manager.
+type Certmanager struct {
```
stlaz, line 10: `CertManager`

hack/lib/init.sh
```diff
@@ -10,6 +10,7 @@ set -o nounset
 set -o pipefail
 
 API_GROUP_VERSIONS="\
+config/v1alpha1 \
```
stlaz, line 13: do we need that backslash in this string?

pkg/operator/interfaces.go
```diff
@@ -1,4 +1,4 @@
-package operatorclient
```
s-urbaniak, line 1: note that this was just to prevent package import clashes with the generated client. i can move it back if this is more idiomatic.

stlaz: I would move it back

pkg/operator/starter.go
```diff
@@ -33,12 +32,18 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
 	}
 
-	genericOperatorClient, dynamicInformers, err := genericoperatorclient.NewClusterScopedOperatorClient(
-		cc.KubeConfig, operatorv1.GroupVersion.WithResource("certmanagers"))
+	certManagerOperatorClient, err := operatorclient.NewForConfig(cc.KubeConfig)
 	if err != nil {
 		return err
 	}
 
+	certManagerInformers := operatorinformers.NewSharedInformerFactory(certManagerOperatorClient, 10*time.Second)
```
stlaz, line 40: resync every 10 seconds seems short, maybe @sttts or @p0lyn0mial might suggest a better value

apis/config/v1alpha1/types_certmanager.go
s-urbaniak: that's actually a good point and needs to be changed in quite some places :+1:

hack/lib/init.sh
s-urbaniak: actyally, no :+1:

pkg/operator/starter.go
s-urbaniak: thank you for the sharp eyes, I actually wanted to set `10*time.Minute`. the dynamic client sets it to 12h (huh) [1] which seems quite large, I'll set to 10 minutes as a first guess.

[1] https://github.com/openshift/library-go/blob/7aab941b4a2b60dc5687744944789456d7a4eb4e/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L33

pkg/operator/interfaces.go
s-urbaniak: ack, ok

pkg/operator/starter.go
```diff
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"github.com/openshift/cert-manager-operator/pkg/operator/operatorclient"
```
stlaz, line 5: order!

pkg/operator/starter.go
```diff
@@ -10,16 +11,15 @@ import (
...
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
-	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"github.com/openshift/cert-manager-operator/pkg/controller/deployment"
-	"github.com/openshift/cert-manager-operator/pkg/operator/operatorclient"
+	cert_manager_operator_client "github.com/openshift/cert-manager-operator/pkg/operator/clientset/versioned"
```
stlaz, line 21: `certmanoperatorclient` maybe?

hack/lib/init.sh
```diff
@@ -10,8 +10,8 @@ set -o nounset
 set -o pipefail
 
 API_GROUP_VERSIONS="\
```
stlaz, line 12: should we remove this backslash, too?

pkg/operator/starter.go
s-urbaniak: i have to set this up somehow.

pkg/operator/starter.go
s-urbaniak: sounds good

hack/lib/init.sh
s-urbaniak: yes i missed that one

PR #9, opened 2021-06-25
Remove non-existing manifests

slaskawi:
Required for https://github.com/openshift/release/pull/19608

### DIFF
--- a/images/ci/Dockerfile
+++ b/images/ci/Dockerfile
@@ -5,5 +5,3 @@ RUN make build --warn-undefined-variables
 
 FROM registry.ci.openshift.org/ocp/4.9:base
 COPY --from=builder /go/src/github.com/openshift/cert-manager-operator/cert-manager-operator /usr/bin/
-COPY manifests /manifests
-LABEL io.openshift.release.operator=true


images/ci/Dockerfile
```diff
@@ -5,5 +5,4 @@ RUN make build --warn-undefined-variables
 
 FROM registry.ci.openshift.org/ocp/4.9:base
 COPY --from=builder /go/src/github.com/openshift/cert-manager-operator/cert-manager-operator /usr/bin/
-COPY manifests /manifests
 LABEL io.openshift.release.operator=true
```
s-urbaniak, line 8: not sure about this one i think we should remove it as well.

s-urbaniak: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/release/annotations.go#L9-L11

s-urbaniak: https://github.com/openshift/oc/blob/a765590e1b87b014b9d81f9ea534460d6dff73f2/pkg/cli/admin/release/new.go#L73-L82

slaskawi: Removed.

Makefile
```diff
@@ -3,7 +3,6 @@ GO_REQUIRED_MIN_VERSION = 1.16
 # Include the library makefiles
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
-	targets/openshift/bindata.mk \
```
s-urbaniak, line 6: bindata per se might still be necessary for the operator, let's leave it

slaskawi: Ok, then I'll create an empty dir with bindata. Otherwise the verify target fails.

PR #18, opened 2021-09-30
AUTH-33 Operator bundle image

slaskawi:
https://issues.redhat.com/browse/AUTH-33

This Pull Request introduces the bundle image into the OpenShift Cert Manager Operator. The bundle is effectively a simplified version of the [upstream one](https://github.com/jetstack/cert-manager-olm).

image

Highlights:
- The name must be different than the upstream Cert Manager to avoid clash in OLM.
- The OLM installs only an Operator, which in turn installs all the CRDs for Cert Manager. This enables us to have more control over the CRDs we use (e.g. we could do crazy things during migrations that are not possible via OLM).
- In certain places I'm using word placeholder. This will be altered later on.

The easiest way to test it out is to use a command similar to this: `make operator-clean operator-push-bundle operator-run-bundle IMAGE_REGISTRY=docker.io/slaskawi RUNTIME=docker`

Still to be done:
- [x] Adjust permissions to successfully run the operator. This requires roughly a day of work.


Makefile
```diff
@@ -1,7 +1,35 @@
...
+IMAGE_REGISTRY :=registry.svc.ci.openshift.org
+
+BUNDLE_IMAGE_NAME=cert-manager-operator-bundle
+BUNDLE_IMAGE_PATH=$(IMAGE_REGISTRY)/$(BUNDLE_IMAGE_NAME)
+BUNDLE_IMAGE_TAG?=latest
+
+# Detect the OS to set per-OS defaults
+OS_NAME=$(shell uname -s)
+# Container runtime
+ifeq ($(OS_NAME), Linux)
```
slaskawi, line 18: This part is copy--pasted from the Compliance Operator. Is there any better way to do it?

bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -0,0 +1,152 @@
...
+    - name: Documentation
+      url: https://github.com/openshift/cert-manager-operator/blob/master/README.md
+  maintainers:
+    - email: cert-manager-maintainers@googlegroups.com
+      name: The cert-manager maintainers
+  maturity: stable
+  minKubeVersion: 1.16.0
+  provider:
+    name: Red Hat
+    url: aos-auth@redhat.com
```
slaskawi, line 151: Is this email alias ok?

bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -0,0 +1,152 @@
...
+    - security
+    - certificates
+  links:
+    - name: Documentation
+      url: https://github.com/openshift/cert-manager-operator/blob/master/README.md
+  maintainers:
+    - email: cert-manager-maintainers@googlegroups.com
+      name: The cert-manager maintainers
+  maturity: stable
+  minKubeVersion: 1.16.0
```
slaskawi, line 148: Maybe 1.21?

bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -0,0 +1,152 @@
...
+  keywords:
+    - TLS
+    - secrets
+    - security
+    - certificates
+  links:
+    - name: Documentation
+      url: https://github.com/openshift/cert-manager-operator/blob/master/README.md
+  maintainers:
+    - email: cert-manager-maintainers@googlegroups.com
```
slaskawi, line 145: Shall we use our own email alias here?

bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -0,0 +1,152 @@
...
+                          fieldRef:
+                            fieldPath: metadata.name
+                      - name: OPERATOR_NAME
+                        value: cert-manager-operator
+                    image: docker.io/slaskawi/cert-manager-operator:latest
+                    imagePullPolicy: Always
+                    name: cert-manager-operator
+                    resources: { }
+                serviceAccountName: cert-manager-operator
+      permissions:
```
slaskawi, line 68: I'll be extending this list very shortly. It will be much, much (, much...) longer.

bundle/manifests/cert-manager.clusterserviceversion.yaml
s-urbaniak: let's use aos-auth-staff@redhat.com

bundle/manifests/cert-manager.clusterserviceversion.yaml
s-urbaniak: let's align this with openshift, hence 1.22. 4.10 will be 1.23 so we can bump once the rebase is done. we cannot express a dependency to openshift right?

bundle/manifests/cert-manager.clusterserviceversion.yaml
s-urbaniak: yes, this should be auth team.

bundle/manifests/cert-manager.clusterserviceversion.yaml
slaskawi: The OpenShift dependency is defined elsewhere. Here we configure Kube only.

bundle/manifests/cert-manager.clusterserviceversion.yaml
slaskawi: Done

bundle/manifests/cert-manager.clusterserviceversion.yaml
slaskawi: Done

PR #20, opened 2021-11-22
Use temporary CDN repo location

slaskawi:


bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   annotations:
     capabilities: Seamless Upgrades
     categories: Security
-    containerImage: docker.io/slaskawi/cert-manager-operator:latest
+    containerImage: registry.stage.redhat.io/cert-manager/cert-manager-operator:1.5
```
s-urbaniak, line 9: hmm
```
$ docker pull registry.stage.redhat.io/cert-manager/cert-manager-operator:1.5
Error response from daemon: Get "https://registry.stage.redhat.io/v2/": dial tcp: lookup registry.stage.redhat.io: no such host
```

s-urbaniak: does this only work via vpn?

PR #22, opened 2021-11-24
Use RELATED_IMAGE_CERT_MANAGER variable

slaskawi:
https://issues.redhat.com/browse/AUTH-108

This PR provides the support for overriding the operand image by using the `RELATED_IMAGE_CERT_MANAGER` environmental variable.

This feature is required for:
- Flexibility in testing
- Supporting SHA Digest pinning and supporting the offline installation mode
- Respinning the images in case of CVE

For more information, please refer to the [OSBS documentation](https://osbs.readthedocs.io/en/latest/users.html#creating-the-relatedimages-section).

### DIFF
--- a/bundle/manifests/cert-manager.clusterserviceversion.yaml
+++ b/bundle/manifests/cert-manager.clusterserviceversion.yaml
@@ -60,6 +60,8 @@ spec:
                             fieldPath: metadata.name
                       - name: OPERATOR_NAME
                         value: cert-manager-operator
+                      - name: RELATED_IMAGE_CERT_MANAGER
+                        value: registry-proxy.engineering.redhat.com/rh-osbs/cert-manager-cert-manager-1.5-rhel-8:latest
                     image: registry-proxy.engineering.redhat.com/rh-osbs/cert-manager-cert-manager-operator-1.5-rhel-8:latest
                     imagePullPolicy: Always
                     name: cert-manager-operator
--- a/pkg/controller/deployment/related_images.go
+++ b/pkg/controller/deployment/related_images.go
@@ -1,0 +1,20 @@
+package deployment
+
+import "os"
+
+const (
+	environmentalVariableCertManagerRelatedImage = "RELATED_IMAGE_CERT_MANAGER"
+)
+
+// Overrides provided image when environmentalVariableCertManagerRelatedImage environment variable is specified.
+// Otherwise, returns provided defaultImage.
+//
+// This function assumes the Cert Manager is provided in all-in-one image. This function will need to be updated
+// is this assumption doesn't hold.
+func certManagerImage(defaultImage string) string {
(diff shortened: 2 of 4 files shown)


bundle/manifests/cert-manager.clusterserviceversion.yaml
```diff
@@ -60,6 +60,8 @@ spec:
                             fieldPath: metadata.name
                       - name: OPERATOR_NAME
                         value: cert-manager-operator
+                      - name: RELATED_IMAGE_CERT_MANAGER
```
soltysh, line 63: Can we keep it consistent with other operators of ours, and just name this `IMAGE` see https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/manifests/0000_25_kube-controller-manager-operator_06_deployment.yaml#L53-L54 or https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_20_kube-apiserver-operator_06_deployment.yaml#L53-L54

slaskawi: The Cert Manager Operator is going to be released thought CPaaS product pipeline (not ART). Having this in mind, we need to use `RELATED_IMAGE_*` variables as described in the link provided in the PR description.

pkg/controller/deployment/related_images.go
```diff
@@ -0,0 +1,20 @@
...
+const (
+	environmentalVariableCertManagerRelatedImage = "RELATED_IMAGE_CERT_MANAGER"
+)
+
+// Overrides provided image when environmentalVariableCertManagerRelatedImage environment variable is specified.
+// Otherwise, returns provided defaultImage.
+//
+// This function assumes the Cert Manager is provided in all-in-one image. This function will need to be updated
+// is this assumption doesn't hold.
+func getCertManagerImage(defaultImage string) string {
```
s-urbaniak, line 14: nit:just use `certManagerImage`, see https://go.dev/doc/effective_go#Getters

pkg/controller/deployment/related_images.go
```diff
@@ -0,0 +1,20 @@
+package deployment
+
+import "os"
+
+const (
+	environmentalVariableCertManagerRelatedImage = "RELATED_IMAGE_CERT_MANAGER"
```
s-urbaniak, line 6: this constant is not used elsewhere, it's fine to inline.

pkg/controller/deployment/related_images.go
slaskawi: Done

pkg/controller/deployment/related_images.go
slaskawi: This variable is quite significant from the prod perspective. I'd like it to stay as `const` so that it screams as you whenever you open this file.

slaskawi:
@s-urbaniak Ready for the next round of review.

PR #28, opened 2021-12-16
Auth 137 move operator to the openshift prefixed namespace

slaskawi:
https://issues.redhat.com/browse/AUTH-137

This Pull Request moves all operands from `cert-manager` into `openshift-cert-manager` namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ verify: verify-scripts verify-codegen-crds
 
 local-deploy-manifests:
 	kubectl apply -f ./manifests
-	kubectl apply -f ./bundle/manifests
+	kubectl apply $$(ls ./bundle/manifests/*.crd.yaml | awk ' { print " -f " $$1 } ')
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -71,7 +71,7 @@ verify: verify-scripts verify-codegen-crds
 
 local-deploy-manifests:
 	kubectl apply -f ./manifests
-	kubectl apply -f ./bundle/manifests
+	kubectl apply $$(ls ./bundle/manifests/*.crd.yaml | awk ' { print " -f " $$1 } ')
```
s-urbaniak, line 74: suggestion: maybe let's put the cluster service version or the CRDs in its own directory?

slaskawi: We'd need to move it around in the downstream then. It's not nice that CRDs and expected to coexist with CSVs in bundles :face_exhaling:

s-urbaniak: ah ... ugh, that's a strange constraint.

PR #29, opened 2021-12-21
Detect other operators

slaskawi:
https://issues.redhat.com/browse/AUTH-136

This Pull Request introduces a mechanism for detecting other operators installed in the cluster. The discovery mechanism scans for predefined `Deployments` (I'm open for better ideas).


pkg/controller/deployment/deployment_checked_adapter.go
```diff
@@ -0,0 +1,41 @@
...
+	//https://github.com/jetstack/cert-manager/tree/master/deploy/charts/cert-manager
+	//https://github.com/IBM/ibm-cert-manager-operator/blob/master/deploy/olm-catalog/ibm-cert-manager-operator/3.17.0/ibm-cert-manager-operator.v3.17.0.clusterserviceversion.yaml
+	"openshift-operators": {"ibm-cert-mana`ger-operator", "cert-manager"},
+	//https://github.com/komish/certmanagerdeployment-operator/blob/main/controllers/componentry/constants.go
+	"cert-manager": {"cert-manager"},
+}
+
+func InitGlobalDeploymentChecker(deploymentGetter v1.DeploymentsGetter) {
+	globalDeploymentChecker = newDeploymentChecker(deploymentGetter)
+}
```
s-urbaniak, line 22: Let's make the deployment checker a regular library-go like controller.

Package level variables are really discouraged in golang, please see https://peter.bourgon.org/go-for-industrial-programming/ and https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html for more background.

Here, concretely, the package level variable is unsynchronized, nothing prevents it from being invoked multiple times, from multiple goroutines which will can cause side effects.

pkg/controller/deployment/deployment_checked_adapter.go
```diff
@@ -0,0 +1,41 @@
...
+
+func GenericDeploymentControllerAdapterFunction(ctx context.Context) (bool, error) {
+	return getReadyStatus(ctx)
+}
+
+func getReadyStatus(ctx context.Context) (bool, error) {
+	deploymentsExist, err := globalDeploymentChecker.isAnyDeploymentPresent(ctx, knownCertManagerDeployments)
+	if err != nil {
+		return false, err
```
s-urbaniak, line 35: here, and elsewhere, let's use `fmt.Errorf("failed to get ready status: %w", err)` to have a proper unwinded error context.

pkg/controller/deployment/deployment_checked_adapter.go
s-urbaniak: i see this is using `StaticResourcesAdapterFunction` as `WithPreconfitionFunc`. This should then still be a struct, and `WithPreconditionsFunc` should reference the struct method.

slaskawi: Let me look into this again, once the library-go PR gets in. 

But I'm not very keen on putting the deployment checker into the library-go. It's essentially a getter (so nothing fancy) with adapters that fit other interfaces.

pkg/controller/deployment/deployment_checked_adapter.go
slaskawi: :+1:

pkg/controller/deployment/deployment_checked_adapter.go
s-urbaniak: I agree on not putting deployment checker in library-go. I am just mandating against global state here.

slaskawi: Got rid of the global state.

slaskawi:
/hold

Needs some more testing. Will unhold when ready.

slaskawi:
/hold

Putting this on hold as we need to agree how to report the status.

slaskawi:
/unhold

Together with @s-urbaniak we decided to go forward with this one.

pkg/controller/deployment/deployment_checker.go
```diff
@@ -0,0 +1,75 @@
...
+	if apierrors.IsNotFound(err) {
+		return false, nil
+	}
+	if err == nil {
+		return true, nil
+	}
+	return false, err
+}
+
+func (d *deploymentChecker) deploymentsGetterImpl(ctx context.Context, namespace, deploymentName string) error {
```
s-urbaniak, line 61: I don't see the value of `deploymentsGetterImpl`. How about we remove it and just do:
```
func (d *deploymentChecker) isDeploymentPresent(ctx context.Context, namespace, deploymentName string) (bool, error) {
	_, err := d.deploymentsGetter.Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
	if apierrors.IsNotFound(err) {
		return false, nil
	}
	if err == nil {
		return true, nil
	}
	return false, err
}
```

pkg/controller/deployment/deployment_checker.go
```diff
@@ -0,0 +1,75 @@
...
+
+var knownCertManagerDeployments = map[string][]string{
+	//https://github.com/jetstack/cert-manager/tree/master/deploy/charts/cert-manager
+	//https://github.com/IBM/ibm-cert-manager-operator/blob/master/deploy/olm-catalog/ibm-cert-manager-operator/3.17.0/ibm-cert-manager-operator.v3.17.0.clusterserviceversion.yaml
+	"openshift-operators": {"ibm-cert-mana`ger-operator", "cert-manager"},
+	//https://github.com/komish/certmanagerdeployment-operator/blob/main/controllers/componentry/constants.go
+	"cert-manager": {"cert-manager"},
+}
+
+type deploymentsGetterFunc func(ctx context.Context, namespace, deploymentName string) error
```
s-urbaniak, line 20: this can be removed, it seems we don't need `deploymentsGetterImpl`.

pkg/controller/deployment/deployment_checker.go
```diff
@@ -0,0 +1,75 @@
...
+
+type deploymentChecker struct {
+	deploymentsGetter     v1.DeploymentsGetter
+	deploymentsGetterFunc deploymentsGetterFunc
```
s-urbaniak, line 24: this could be removed

pkg/controller/deployment/deployment_checker.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func newDeploymentChecker(deploymentsGetter v1.DeploymentsGetter) *deploymentChecker {
+	ret := &deploymentChecker{
+		deploymentsGetter: deploymentsGetter,
+	}
+	ret.deploymentsGetterFunc = ret.deploymentsGetterImpl
```
s-urbaniak, line 31: this could be removed

slaskawi:
Closing this as we decided to push this responsibility to OLM. Once we add CRDs into the Bundle, OLM will automatically detect other owners of the CRDs.

PR #30, opened 2022-01-13
Always create certmanager/cluster CR

slaskawi:
https://issues.redhat.com/browse/AUTH-105

This PR enables the Operator to always create the `certmanager/cluster` CR, so that other pieces of code could assume it's always there. This approach has stronger guarantees than defining the `certmanager/cluster` in the OLM bundle (someone can just delete it). The convention of having those objects comes down from library-go and core. Most of the controllers need it to write their status. 

This decision has a few consequences:
- We no longer need an example in the bundle. 
- We don't need to mention this requirement anywhere in the docs (but I'm doing it either way - to make users aware that there is something like this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ The Operator uses the [upstream deployment manifests](https://github.com/jetstac
 - [cert_manager_controller_deployment.go](pkg/controller/deployment/cert_manager_controller_deployment.go)
 - [cert_manager_webhook_deployment.go](pkg/controller/deployment/cert_manager_webhook_deployment.go)
 
-The deployment is triggered upon creating a cluster-scoped `CertManager` object named `cluster`. An example might be found in the [deploy](deploy) directory.
+The Operator automatically deploys a cluster-scoped `CertManager` object named `cluster` if it's missing (with default values).
 
 ### Directory structure
 
--- a/bundle/manifests/cert-manager.clusterserviceversion.yaml
+++ b/bundle/manifests/cert-manager.clusterserviceversion.yaml
@@ -13,19 +13,6 @@ metadata:
     support: The OpenShift Engineering Team
     operatorframework.io/suggested-namespace: openshift-cert-manager-operator
     operatorframework.io/cluster-monitoring: "true"
-    alm-examples: |
-      [
-        {
-          "apiVersion": "operator.openshift.io/v1alpha1",
-          "kind": "CertManager",
-          "metadata": {
-            "name": "cluster"
-          },
-          "spec": {
-            "managementState": "Managed"
-          }
-        }
-      ]
   name: openshift-cert-manager.v0.0.1
   namespace: placeholder
 spec:
--- a/pkg/controller/deployment/default_cert_manager_controller.go
+++ b/pkg/controller/deployment/default_cert_manager_controller.go
@@ -1,0 +1,59 @@
+package deployment
+
+import (
+	"context"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/cert-manager-operator/apis/operator/v1alpha1"
+	alpha1 "github.com/openshift/cert-manager-operator/pkg/operator/clientset/versioned/typed/operator/v1alpha1"
+)
+
+type DefaultCertManagerController struct {
+	operatorClient    v1helpers.OperatorClient
+	controllerFactory *factory.Factory
+	recorder          events.Recorder
+	certManagerClient alpha1.OperatorV1alpha1Interface
+}
+
+func NewDefaultCertManagerController(operatorClient v1helpers.OperatorClient, certManagerClient alpha1.OperatorV1alpha1Interface, eventsRecorder events.Recorder) factory.Controller {
+	controller := DefaultCertManagerController{
+		operatorClient:    operatorClient,
+		certManagerClient: certManagerClient,
+		controllerFactory: factory.New().ResyncEvery(time.Minute).WithInformers(
+			operatorClient.Informer(),
+		),
+		recorder: eventsRecorder.WithComponentSuffix("default-cert-manager-controller"),
+	}
+
+	return controller.controllerFactory.WithSync(controller.sync).ToController("DefaultCertManager", controller.recorder)
+}
+
+func (c *DefaultCertManagerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	_, _, _, err := c.operatorClient.GetOperatorState()
+	if apierrors.IsNotFound(err) {
+		syncCtx.Recorder().Eventf("StatusNotFound", "Creating \"cluster\" certmanager")
+		_, err = c.createDefaultCertManager(ctx)
+	}
+	return err
+}
+
+func (c *DefaultCertManagerController) createDefaultCertManager(ctx context.Context) (*v1alpha1.CertManager, error) {
+	cm := &v1alpha1.CertManager{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+		Spec: v1alpha1.CertManagerSpec{
+			OperatorSpec: operatorv1.OperatorSpec{
+				ManagementState: operatorv1.Managed,
+			},
+		},
+	}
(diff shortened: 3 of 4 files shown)


pkg/operator/operatorclient/operatorclient.go
```diff
@@ -34,12 +36,17 @@ func (c OperatorClient) Informer() cache.SharedIndexInformer {
 	return c.Informers.Operator().V1alpha1().CertManagers().Informer()
 }
 
+// GetOperatorState returns the Operator state. If it's missing (e.g. the certmanager CR hasn't been creates),
```
s-urbaniak, line 39: nit: let's leave out the brackets

pkg/operator/operatorclient/operatorclient.go
```diff
@@ -34,12 +36,17 @@ func (c OperatorClient) Informer() cache.SharedIndexInformer {
...
+// it will create a default one.
 func (c OperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
 	instance, err := c.Informers.Operator().V1alpha1().CertManagers().Lister().Get("cluster")
+	if apierrors.IsNotFound(err) {
+		//TODO: Consider using propagated context here but this requires changes in library-go. Maybe it's not worth it?
+		instance, err = c.createDefaultCertManager(context.TODO())
```
s-urbaniak, line 45: As discussed I believe we have the open question how this can be removed at uninstall time.

pkg/operator/operatorclient/operatorclient.go
```diff
@@ -34,12 +36,17 @@ func (c OperatorClient) Informer() cache.SharedIndexInformer {
 	return c.Informers.Operator().V1alpha1().CertManagers().Informer()
 }
 
+// GetOperatorState returns the Operator state. If it's missing (e.g. the certmanager CR hasn't been creates),
+// it will create a default one.
 func (c OperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
```
s-urbaniak, line 41: It seems strange to me that `GetOperator` state has side effects. Do we have a better place?

pkg/operator/operatorclient/operatorclient.go
slaskawi: If you're uninstalling the Operator? I think it will be removed by OLM, but let me double check this scenario.

pkg/operator/operatorclient/operatorclient.go
slaskawi: Yeah... thinking about this a bit more it's weird. Let me introduce a new Controller to ensure it's always there. It will be cleanest solution I think.

pkg/operator/operatorclient/operatorclient.go
slaskawi: I'm not touching this file anymore.

pkg/operator/operatorclient/operatorclient.go
slaskawi: Done

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,67 @@
...
+		controllerFactory: factory.New().ResyncEvery(time.Minute).WithInformers(
+			operatorClient.Informer(),
+		),
+		recorder: eventsRecorder.WithComponentSuffix("default-cert-manager-controller"),
+		name:     controllerName,
+	}
+}
+
+func (c *DefaultCertManagerController) Run(ctx context.Context, workers int) {
+	c.controllerFactory.WithSync(c.Sync).ToController("DefaultCertManager_"+c.name, c.recorder).Run(ctx, workers)
```
s-urbaniak, line 43: nit: you want to return the Controller in the `NewDefaultCertManagerController` constructor and remove this method, i.e.
```
	c := &DefaultCertManagerController{
		operatorClient:    operatorClient,
		certManagerClient: certManagerClient,
		controllerFactory: factory.New().ResyncEvery(time.Minute).WithInformers(
			operatorClient.Informer(),
		),
		recorder: eventsRecorder.WithComponentSuffix("default-cert-manager-controller"),
		name:     controllerName,
	}
	return c.controllerFactory.WithSync(c.Sync).ToController("DefaultCertManager_"+c.name, c.recorder)
```

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (c *DefaultCertManagerController) Sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
s-urbaniak, line 46: let's not make this method public. the controller factory `Sync` is.

pkg/controller/deployment/default_cert_manager_creator.go
slaskawi: Fixed.

pkg/controller/deployment/default_cert_manager_creator.go
slaskawi: Fixed.

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,65 @@
...
+		operatorClient:    operatorClient,
+		certManagerClient: certManagerClient,
+		controllerFactory: factory.New().ResyncEvery(time.Minute).WithInformers(
+			operatorClient.Informer(),
+		),
+		recorder: eventsRecorder.WithComponentSuffix("default-cert-manager-controller"),
+		name:     controllerName,
+	}
+
+	return controller.controllerFactory.WithSync(controller.sync).ToController("DefaultCertManager_"+controller.name, controller.recorder)
```
s-urbaniak, line 41: let's stay consistent with CamelCase

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,65 @@
...
+
+type DefaultCertManagerController struct {
+	operatorClient    v1helpers.OperatorClient
+	controllerFactory *factory.Factory
+	recorder          events.Recorder
+	name              string
+	certManagerClient alpha1.OperatorV1alpha1Interface
+}
+
+func (c *DefaultCertManagerController) Name() string {
```
s-urbaniak, line 26: curious: why is this needed?

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,65 @@
...
+	return c.name
+}
+
+func NewDefaultCertManagerController(controllerName string, operatorClient v1helpers.OperatorClient, certManagerClient alpha1.OperatorV1alpha1Interface, eventsRecorder events.Recorder) factory.Controller {
```
s-urbaniak, line 30: I'd vote for removing `controllerName`.

pkg/controller/deployment/default_cert_manager_creator.go
s-urbaniak: Actually, I suggest just to leave `DefaultCertManager`.

pkg/controller/deployment/default_cert_manager_creator.go
```diff
@@ -0,0 +1,65 @@
...
+	controller := DefaultCertManagerController{
+		operatorClient:    operatorClient,
+		certManagerClient: certManagerClient,
+		controllerFactory: factory.New().ResyncEvery(time.Minute).WithInformers(
+			operatorClient.Informer(),
+		),
+		recorder: eventsRecorder.WithComponentSuffix("default-cert-manager-controller"),
```
s-urbaniak, line 37: I suggest to just leave `default-cert-manager`.

pkg/controller/deployment/default_cert_manager_creator.go
slaskawi: Fixed.

pkg/controller/deployment/default_cert_manager_creator.go
slaskawi: Required by the `factory.Controller` interface.

pkg/controller/deployment/default_cert_manager_creator.go
slaskawi: Removed.

pkg/controller/deployment/default_cert_manager_creator.go
s-urbaniak: this is returning an instance created by the controller factory, `Name()` is implemented by `baseController`. The `Name()` function here has no effect.

s-urbaniak: https://github.com/openshift/cert-manager-operator/blob/b9570ecc15835f9e3b266957f0e0be01b97e1a09/vendor/github.com/openshift/library-go/pkg/controller/factory/base_controller.go#L45-L47

slaskawi: Removed

PR #31, opened 2022-01-26
Move CRDs into the bundle

slaskawi:
This PR moves CRDs from the operator into the bundle. 

Unfortunately full e2e testing will be done downstream.


jsonnet/main.jsonnet
```diff
@@ -79,7 +79,7 @@ local suffix = {
 // create a path including the file name based on the item.
 local path(item) =
   // CRDs go into cert-manager-crds directory
-  if item.kind == 'CustomResourceDefinition' then 'cert-manager-crds/' + item.metadata.name + '-' + suffix[item.kind] + '.yaml'
+  if item.kind == 'CustomResourceDefinition' then '../bundle/manifests/' + item.metadata.name + '-' + suffix[item.kind] + '.yaml'
```
s-urbaniak, line 82: is it hard to switch to the root directory instead of it being `bindata`. we could then write `bundle/manifests/` for the crd generation target and `bindata/cert-manager-deployment/` and `bindata/cert-manager-deployment/` for the deployment manifests.

s-urbaniak: it would make it more explicit here.

slaskawi: Sure, done.

PR #37, opened 2022-02-02
Bundle updates

slaskawi:
This Pull Request introduces the following changes in the bundle:

- Add permissions to leases
- Use tech-preview instead of dev-preview
- Use the operand version

### DIFF
--- a/pkg/controller/deployment/common.go
+++ b/pkg/controller/deployment/common.go
@@ -1,13 +1,7 @@
 package deployment
 
-import (
-	"os"
-)
-
 const (
 	operatorName      = "cert-manager"
 	operandNamePrefix = ""
 	conditionsPrefix  = "CertManager"
 )
-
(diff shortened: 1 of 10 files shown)


pkg/controller/deployment/common.go
```diff
@@ -10,4 +10,11 @@ const (
...
 )
 
-var targetVersion = os.Getenv("OPERAND_CERT_MANAGER_IMAGE_VERSION")
+var targetVersion = "dev"
+
+func init() {
+	version := os.Getenv("OPERATOR_VERSION")
+	if len(version) != 0 {
+		targetVersion = version
+	}
```
s-urbaniak, line 19: let's migrate this to the startup methods and dependency inject. having mutable package level state can cause issues, see https://peter.bourgon.org/go-for-industrial-programming/

PR #42, opened 2022-02-04
Bundle adjustments

slaskawi:
This Pull Request introduces the following changes:

- Always delete operands
- Add monitoring annotation to the operand


bindata/cert-manager-deployment/cert-manager-namespace.yaml
```diff
@@ -1,4 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  annotations:
+    openshift.io/cluster-monitoring: "true"
```
s-urbaniak, line 5: how about also monitoring the operator namespace?

slaskawi: It's already done here: https://github.com/openshift/cert-manager-operator/pull/42/files#diff-5c0b7320c5f7cb195262ffd7f9f87191abda01a3d7563d2e9588ce5e0638825eR15

Note, that when creating an operator via UI, the user might opt-out from this feature. But there's nothing we can do about this.

PR #63, opened 2022-04-18
Added owners for e2e tests

y4sht:
Added @xingxingxia @y4sht as reviewers and approvers for e2e tests

### DIFF
--- a/test/OWNERS.md
+++ b/test/OWNERS.md
@@ -1,0 +1,10 @@
+reviewers:
+  - s-urbaniak
+  - xingxingxia
+  - slaskawi
+  - y4sht
+approvers:
+  - s-urbaniak 
+  - xingxingxia
+  - slaskawi
+  - y4sht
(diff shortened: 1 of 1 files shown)


test/OWNERS.md
```diff
@@ -0,0 +1,10 @@
+reviewers:
+  - s-urbaniak
+  - xingxingxia
+  - slaskawi
+  - y4sht
+approvers:
+  - s-urbaniak 
+  - xingxingxia
+  - slaskawi
+  - y4sht
```
xingxingxia, line 10: Even if it is only for test directory, maybe I (xingxingxia) and y4sht are better to only be reviewers, at least at the current moment and in future after we exercise with more tests merged we can upgrade to approvers if slaskawi would think fine :) , anyway, just a chime-in.

slaskawi:
I'm leaving this up to @s-urbaniak 

/assign @s-urbaniak

PR #64, opened 2022-04-27
Added files for running tests in CVP

y4sht:


cvp.yaml
```diff
@@ -0,0 +1,13 @@
+---
+tests:
+  playbooks:
+    - location: test/e2e-cert-manager.yaml
+results:
+  delivery_types:
+    email:
+      - slaskawi@redhat.com
+      - xxia@redhat.com
+      - ytripath@redhat.com
```
slaskawi, line 10: Probably worth adding @s-urbaniak 's email as well.

slaskawi (changes requested):
Great to see progress on this one! 

Just one simple fix and we should be good to merge this.

cvp.yaml
y4sht: +1

y4sht:
@slaskawi Updated the PR with @s-urbaniak's email PTAL

Makefile
```diff
@@ -1,5 +1,5 @@
 GO_REQUIRED_MIN_VERSION = 1.17
-GO_TEST_FLAGS=-v
+# GO_TEST_FLAGS=-v
```
xingxingxia, line 2: Just minor nag, why comment out?

y4sht: Thanks for the catch, missed in self review, I meant to remove it, so that it does not give verbose flags for unit tests written by @slaskawi 
@slaskawi shall I keep it or remove it, since now our e2e tests are not technically tests, they compile to a binary, so I do not have any need for this

slaskawi: I'm OK with removing it.

Makefile
```diff
@@ -8,6 +8,7 @@ IMAGE_ORG?=openshift-cert-manager
 IMAGE_TAG?=latest
 IMAGE_OPERATOR?=$(IMAGE_REGISTRY)/$(IMAGE_ORG)/cert-manager-operator:$(IMAGE_TAG)
 IMAGE_OPERATOR_BUNDLE?=$(IMAGE_REGISTRY)/$(IMAGE_ORG)/cert-manager-operator-bundle:$(IMAGE_TAG)
+TESTS_IMAGE?=$(IMAGE_REGISTRY)/$(IMAGE_ORG)/cert-manager-scorecard-tests:$(IMAGE_TAG)
```
slaskawi, line 11: Can we preserve the image naming convention - `cert-manager-operator-scorecard-test`?

Makefile
```diff
@@ -44,12 +45,31 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
 $(call build-image,cert-manager-operator,$(IMAGE_OPERATOR),./images/ci/Dockerfile,.)
 $(call build-image,cert-manager-operator-bundle,$(IMAGE_OPERATOR_BUNDLE),./bundle/bundle.Dockerfile,./bundle)
 
+# generate scorecard image targets
+.PHONY: build/custom-scorecard-tests
+build/custom-scorecard-tests:
+	go build -mod=vendor $(GO_GCFLAGS) $(GO_ASMFLAGS) -o $(BUILD_DIR)/$(@F) ./test
```
slaskawi, line 51: What does the `$(@F)` do?

Makefile
```diff
@@ -44,12 +45,31 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+
+.PHONY: build-e2e-image
+build-e2e-image:
+	docker build -f ./images/custom-scorecard-tests/Dockerfile ./ -t quay.io/ytripath/cert-manager-operator-e2e-tests:latest
```
slaskawi, line 55: We should parametrize image coordinates

Makefile
```diff
@@ -44,12 +45,31 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+	docker push quay.io/ytripath/cert-manager-operator-e2e-tests:latest
+
+.PHONY: run-e2e-image
+run-e2e-image: test-e2e-wait-for-stable-state
+	oc apply -f test/rbac.yaml
+	operator-sdk scorecard ./bundle --namespace=cm-tests --wait-time=600s -o json
```
slaskawi, line 61: Namespace and timeouts should be parametrized.

bundle/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,29 @@
+kind: Configuration
+apiversion: scorecard.operatorframework.io/v1alpha3
+metadata:
+  name: config
+
+stages:
+  - tests:
+      - image: quay.io/ytripath/cert-manager-operator-e2e-tests:latest
```
slaskawi, line 8: This should be parametrized. Maybe Kustomize could help here?

images/custom-scorecard-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Build the custom-scorecard-tests binary
+FROM --platform=x86_64 golang:1.17.9 as builder
```
slaskawi, line 2: Can we use the same images as in `/ci` directory? Unification could be a good thing here.

images/custom-scorecard-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ARG TARGETARCH
```
slaskawi, line 3: Cert Manager Operator is x86 ATM. @s-urbaniak Do we want to add more architectures?

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+	cfg, err := apimanifests.GetBundleFromDir(PodBundleRoot)
+	fmt.Println(err)
+	if err != nil {
+		log.Fatalf("Failed to read bundle")
+		fmt.Printf("%s\n", logBuf.String())
+	}
+
+	var result scapiv1alpha3.TestResult
+
+	switch entrypoint[0] {
```
slaskawi, line 65: I'm not sure I like this approach. Whenever we add a new test, we'd need to update this switch. Can we somehow make it more robust?

slaskawi (changes requested):
The direction is good I think. What I also miss from this Pull Request is an instruction telling the developers how to run this locally.

Makefile
y4sht: It is an automatic variable which picks up `custom-scorecard-tests` automatically from the target name `build/custom-scorecard-tests`

Makefile
y4sht: Do you mean changing `latest` to `$(IMAGE_TAG)`

images/custom-scorecard-tests/Dockerfile
y4sht: +1

bundle/tests/scorecard/config.yaml
y4sht: @slaskawi Would it be okay, if I added configs to generate this file when `make bundle` is run, using `kustomize`

Makefile
```diff
@@ -44,12 +45,31 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+	oc delete -f test/rbac.yaml
+
 # exclude e2e test from unit tests
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
 
 # re-use test-unit target for e2e tests
```
xingxingxia, line 67: This comment should be removed given you redefine test-e2e

Makefile
y4sht: This target is just used for local testing, also, we are setting up rbac in the `test/rbac.yaml` file for the tests to run, which includes the namespace, parameterizing the namescape will make this complex

y4sht:
> The direction is good I think. What I also miss from this Pull Request is an instruction telling the developers how to run this locally.

Sure, I'll add that, it's
`make test-e2e` to build and run tests binary
`make run-e2e-image` to run the tests from the docker image

Makefile
y4sht: substituted with oc commands

Makefile
y4sht: Changed image coordinates to variables

Makefile
slaskawi: Nice hack!

Makefile
```diff
@@ -44,12 +45,31 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
-test-e2e: test-e2e-wait-for-stable-state
-	$(MAKE) GO_TEST_PACKAGES=./test/e2e/... test-unit
+test-e2e:
+	go build -o ./e2e-tests test/main.go
+	./e2e-tests TestSelfSignedCerts ./bundle
+	./e2e-tests TestACMECertsIngress ./bundle
+	./e2e-tests TestCertRenewal ./bundle
```
xingxingxia, line 72: Same as https://github.com/openshift/cert-manager-operator/pull/64/files#r871078707 , if 20 cases are added, current approach will make it have to enumerate 20 case names.

Makefile
xingxingxia: Per convention:
```
Usage:  docker build [OPTIONS] PATH | URL | -
```
It is better to put your `./` at the end.

images/custom-scorecard-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Build
+RUN GOOS=linux GOARCH=$TARGETARCH BUILD_DIR=/workspace make GO_REQUIRED_MIN_VERSION:= build/custom-scorecard-tests
+
+# Final image.
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
+
+COPY --from=builder /workspace/custom-scorecard-tests /usr/local/bin/custom-scorecard-tests
+
+ENTRYPOINT ["/usr/local/bin/custom-scorecard-tests"]
```
xingxingxia, line 18: Viewing from "https://github.com/openshift/cert-manager-operator/pull/64/files" page, it does not have a newline character.

images/custom-scorecard-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /workspace
+
+# Copy the go source
+COPY . .
+
+# Build
+RUN GOOS=linux GOARCH=$TARGETARCH BUILD_DIR=/workspace make GO_REQUIRED_MIN_VERSION:= build/custom-scorecard-tests
+
+# Final image.
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
```
xingxingxia, line 14: Is "latest" tag (https://catalog.redhat.com/software/containers/ubi8-minimal/5c64772edd19c77a158ea216) fine? If fine, how about using "latest"?

images/custom-scorecard-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM --platform=x86_64 golang:1.17.9 as builder
+ARG TARGETARCH
+
+WORKDIR /workspace
+
+# Copy the go source
+COPY . .
+
+# Build
+RUN GOOS=linux GOARCH=$TARGETARCH BUILD_DIR=/workspace make GO_REQUIRED_MIN_VERSION:= build/custom-scorecard-tests
```
xingxingxia, line 11: The "FROM" image should be linux, so is GOOS needed?
Is GO_REQUIRED_MIN_VERSION:= needed?

bundle/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,29 @@
...
+        entrypoint:
+          - custom-scorecard-tests
+          - TestACMECertsIngress
+        labels:
+          suite: custom
+          test: TestACMECertsIngress
+      - image: quay.io/ytripath/cert-manager-operator-e2e-tests:latest
+        entrypoint:
+          - custom-scorecard-tests
+          - TestCertRenewal
```
xingxingxia, line 25: Same question as https://github.com/openshift/cert-manager-operator/pull/64/files#r871078707 . After https://github.com/openshift/cert-manager-operator/pull/64/files#r871078707 is solved, maybe we need not enumerate scenarios under "- tests:" as well and only one single item under "- tests:" is enough?

test/rbac.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: cert-manager-operator-e2e
+subjects:
+  - kind: ServiceAccount
+    name: cert-manager-operator-e2e-test-runner
+    namespace: cm-tests
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
xingxingxia, line 25: I'm not sure what is the biggest permissions are needed. I suggest code uses least needed permissions. For example, if the "admin" role on namespace cm-tests for the SA is enough, I suggest using only "admin" role on namespace cm-tests.

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/openshift/cert-manager-operator/test/e2e"
+	lib "github.com/openshift/cert-manager-operator/test/library"
+	scapiv1alpha3 "github.com/operator-framework/api/pkg/apis/scorecard/v1alpha3"
+	apimanifests "github.com/operator-framework/api/pkg/manifests"
+	"k8s.io/klog/v2"
+
+	"log"
+)
+
+var PodBundleRoot = "/bundle"
```
xingxingxia, line 33: Is initial lower-case podBundleRoot fine? Upper-case initial var means usable in other packages.

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+func main() {
+	entrypoint := os.Args[1:]
```
xingxingxia, line 36: Maybe plural endpoints?

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+	if len(entrypoint) == 2 {
+		PodBundleRoot = entrypoint[1]
+	}
+	// Disable klog for this test
```
xingxingxia, line 40: It is better to add a bit explanation in the comment :)

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+	flag.Set("alsologtostderr", "false")
+	flag.Parse()
+
+	// Output logs to a buffer instead of stdout.
+	var logBuf bytes.Buffer
+	log.SetOutput(&logBuf)
+
+	if len(entrypoint) == 0 {
+		log.Fatalf("Test name argument is required")
+		fmt.Printf("%s\n", logBuf.String())
```
xingxingxia, line 52: Would it be empty which does not need Printf?

test/main.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Read the pod's untar'd bundle from a well-known path.
+	cfg, err := apimanifests.GetBundleFromDir(PodBundleRoot)
+	fmt.Println(err)
+	if err != nil {
+		log.Fatalf("Failed to read bundle")
+		fmt.Printf("%s\n", logBuf.String())
```
xingxingxia, line 60: fmt.Println(err) should be moved within the if block?

https://pkg.go.dev/log#Fatalf says:
Fatalf is equivalent to Printf() followed by a call to os.Exit(1)
So Fatalf should be moved after line 60?

test/main.go
y4sht: Updated with a reflector that can call functions by parameter name

test/rbac.yaml
y4sht: Since we need to manage pods in another namespace, create ClusterIssuers and other non namespaced resources, I went  with using cluster-admin, without it ClusterIssuer creation was failing

images/custom-scorecard-tests/Dockerfile
y4sht: I don't think they are needed, will remove them

Makefile
y4sht: I intend to remove this target, this only tests the binary, `make run-e2e-tests` actually runs the docker image

images/custom-scorecard-tests/Dockerfile
y4sht: not using this image anymore

Makefile
```diff
@@ -44,12 +50,34 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+
 # exclude e2e test from unit tests
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
 
-# re-use test-unit target for e2e tests
-test-e2e: test-e2e-wait-for-stable-state
-	$(MAKE) GO_TEST_PACKAGES=./test/e2e/... test-unit
+test-e2e:
+	go build -o ./e2e-tests test/main.go
+	./e2e-tests --bundle ./bundle all
```
xingxingxia, line 80: [./bundle](https://github.com/openshift/cert-manager-operator/tree/master/bundle) is originally used for the the product bundle to be released, is it good to mix the test bundle under the product bundle? Can we not put test bundle stuff under product bundle dir?

Makefile
```diff
@@ -44,12 +50,34 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+
+.PHONY: push-e2e-image
+push-e2e-image: build-e2e-image
+	docker push $(TESTS_IMAGE)
+
+.PHONY: run-e2e-image
+run-e2e-image:
+	IMAGE=${TESTS_IMAGE} envsubst < bundle/tests/scorecard/base.yaml > bundle/tests/scorecard/config.yaml 
+	oc new-project $(TESTS_NAMESPACE) || oc project $(TESTS_NAMESPACE)
+	oc create sa $(TESTS_SA) -n $(TESTS_NAMESPACE) || oc get sa $(TESTS_SA) -n $(TESTS_NAMESPACE)
```
xingxingxia, line 70: Minor issue, maybe below will read more straightforward
```
oc project $(TESTS_NAMESPACE) || oc new-project $(TESTS_NAMESPACE)
oc get sa $(TESTS_SA) -n $(TESTS_NAMESPACE) || oc create sa $(TESTS_SA) -n $(TESTS_NAMESPACE)
```

Makefile
```diff
@@ -44,12 +50,34 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+	oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:$(TESTS_NAMESPACE):$(TESTS_SA)
+	operator-sdk scorecard ./bundle --namespace=$(TESTS_NAMESPACE) --wait-time=$(TESTS_WAIT_TIME) -o json || oc delete project $(TESTS_NAMESPACE)
+	oc delete project $(TESTS_NAMESPACE)
```
xingxingxia, line 73: Minor issue, line 72 maybe could use `true` given line 73 exists anyway.
```
operator-sdk scorecard ./bundle --namespace=$(TESTS_NAMESPACE) --wait-time=$(TESTS_WAIT_TIME) -o json || true
```

test/e2e/cert_manager_deployment.go
```diff
@@ -119,36 +154,55 @@ func TestACMECertsIngress(t *testing.T) {
 	}
 	ingress, err = loader.KubeClient.NetworkingV1().Ingresses(ingress.ObjectMeta.Namespace).Create(ctx, ingress, metav1.CreateOptions{})
 	require.NoError(t, err)
+	if !assert.NoError(t, err) {
```
xingxingxia, line 157:
```
$ grep -nr 'require\.'
...
test/library/dynamic_resources.go:77:		require.NotEmpty(d.t, unstructuredObj.GetNamespace(), "Namespace can not be empty!")
test/library/kube_client.go:20:	require.NoError(t, err)
test/library/kube_client.go:22:	require.NoError(t, err)
test/library/kube_client.go:24:	require.NoError(t, err)
test/library/kube_client.go:41:	require.NoError(t, err)
test/e2e/cert_manager_deployment.go:156:	require.NoError(t, err)
```
The "grep" still finds some "require." occurrences, need all be removed?

Makefile
```diff
@@ -44,12 +50,34 @@ $(call add-bindata,assets,./bindata/...,bindata,assets,pkg/operator/assets/binda
...
+
 # exclude e2e test from unit tests
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
 
-# re-use test-unit target for e2e tests
-test-e2e: test-e2e-wait-for-stable-state
-	$(MAKE) GO_TEST_PACKAGES=./test/e2e/... test-unit
+test-e2e:
+	go build -o ./e2e-tests test/main.go
```
xingxingxia, line 79: The binary "e2e-tests" and binary "custom-scorecard-tests" have same functions, right? If yes, would it be somewhat strange if we use two binary names for just same program? How about removing line 79 and just appending build/custom-scorecard-tests to line 78 "test-e2e: "?

ibihim (changes requested):
Why did you stop using the Go native library testing framework and create a binary?

If you never used the Go testing library, the documentation is pretty good: https://pkg.go.dev/testing.
Otherwise there is a repository with examples and follow up links: https://github.com/ardanlabs/gotraining/tree/master/topics/go/testing/tests

If you need help or want to have a discussion, ping me on Slack, continue here or lets schedule a meeting.

y4sht:
> Why did you stop using the Go native library testing framework and create a binary?
> 
> If you never used the Go testing library, the documentation is pretty good: https://pkg.go.dev/testing. Otherwise there is a repository with examples and follow up links: https://github.com/ardanlabs/gotraining/tree/master/topics/go/testing/tests
> 
> If you need help or want to have a discussion (maybe I am lacking background knowledge and you help me understand), ping me on Slack, continue here or lets schedule a meeting.

Scorecard is expecting tests to be present as a binary which can handle entrypoints. That binary is then containerized and scorecard uses the image to run the tests https://sdk.operatorframework.io/docs/testing-operators/scorecard/custom-tests/
We had the implementation as tests, but I had to switch to this even though writing and running it is a bit awkward, I agree that implementing this using the testing library was a far cleaner option.

ibihim:
@y4sht, when I run `go help test`, I get those interesting flags:

```
usage: go test [build/test flags] [packages] [build/test flags & test binary flags]

// Left out some description

In addition to the build flags, the flags handled by 'go test' itself are:

	-args
	    Pass the remainder of the command line (everything after -args)
	    to the test binary, uninterpreted and unchanged.
	    Because this flag consumes the remainder of the command line,
	    the package list (if present) must appear before this flag.

	-c
	    Compile the test binary to pkg.test but do not run it
	    (where pkg is the last element of the package's import path).
	    The file name can be changed with the -o flag.
	
// left out some flags

	-o file
	    Compile the test binary to the named file.
	    The test still runs (unless -c or -i is specified).

The test binary also accepts flags that control execution of the test; these
flags are also accessible by 'go test'. See 'go help testflag' for details.
```

The flags `-args` and `-c` seem like what you would need to create a test binary with flags or as an entry point for a docker image.

PR #68, opened 2022-06-14
If you set unsupportedConfigOverrides you replace all the controller args

rarguello:
Because we replace the args in here:

https://github.com/openshift/cert-manager-operator/blob/master/pkg/controller/deployment/unsupported_config_overrides.go#L20

The cluster-resource-namespace will not be set, it will get the default value (cert-manager) and you won't find the Secrets for your ClusterIssuers.

Maybe we need add args instead of replacing them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ spec:
     # Here's an example to supply custom DNS settings.
     controller:
       args:
-        - "--dns01-recursive-nameservers=1.1.1.1:53"
-        - "--dns01-recursive-nameservers-only"
+        - --v=2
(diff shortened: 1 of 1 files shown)


kenmoini:
Hey @rarguello - seems like this is by design, otherwise you wouldn't be able to override the defaults such as the verbosity flag.

Adding logic to check override input vs the defaults can cause complexity that may become brittle and difficult to maintain in the long term.

Your README contribution looks good though, there certainly needs to be additional guidance for the arg overrides.

Thoughts?

swghosh:
/close
We've a new PR #87 for implementing the union of default arg values with user provided unsupported override args to handle this problem.

PR #83, opened 2022-11-21
CFE-654: Feature/olm scaffolding

thejasn:
## Description
---
Migration of cert manager operator to use `operator-sdk` for OLM bundle generation and packaging. It changes how bundles are bootstrapped and tweaked. 

OLM scaffolding generated using,
```go
$> operator-sdk init --domain openshift.io --project-name cert-manager-operator --repo github.com/openshift/cert-manager-operator
$> operator-sdk create api --group operator.openshift.io --kind CertManager --namespaced=false --version v1alpha1
```

## Notable Changes
- Operator namespace changed from `openshift-cert-manager-operator` -> `cert-manager-operator`.
- Operand namespace changed from `openshift-cert-manager` -> `cert-manager`.
- `config/alphav1` API dropped for the operator.
- Bundle version changed to `0.0.1` (default).
- New `Dockerfile` introduced for CI.

## Change Log
- `apis` -> `api`
    - Renamed since `config/alphav1` API was removed.
- `bundle` : Regenerated using `operator-sdk`.
    - Operator CSV updated
    - Operator rbac regenerated using kubebuilder annotations
    - Operator install mode changed from `AllNamespaces` to `SingleNamespace` and `OwnedNamespace`.
- `cmd` : Deleted and `main.go` moved to project root.
- `config` : Bundle template generated using `operator-sdk`.
- `hack`
    - `hack/update-cert-manager-manifests.sh` : Update manifest destination as `config/crd/bases/`.
    - `hack/update-clientgen.sh` and `hack/update-deepcopy.sh` : Update `api` file path.
- `jsonnet` : Update script with the new namespaces.
- `pkg/*`
    - `pkg/cmd/operator/cmd.go` : Update operator namespace
    - `pkg/controller/certmanager_controller.go` : Introduced placeholder controller along with required kubebuilder annotations.
    - `pkg/operator` : Re-generate operator client and helper functions.

PR #84, opened 2022-12-06
CFE-676: Feature/decouple cert manager operator as a clusteroperator

thejasn:
Description
---
Decouples `cert-manager-operator` from registering as a clusteroperator. After this change the operator will not show up in the list of `ClusterOperators` on the console. 

Verified on cluster by installing bundle.
image


thejasn:
Wait for rebase post #83 being merged.

pkg/controller/deployment/generic_deployment_controller.go
```diff
@@ -22,6 +22,13 @@ import (
 	"github.com/openshift/cert-manager-operator/pkg/operator/operatorclient"
 )
 
+// TODO: This is just a temporary placeholder.
```
TrilokGeer, line 25: In this PR, client registration is removed. Am I right to understand, the status can be seen in implementation of CFE-668?

thejasn: Status of deployments (3 controllers from upstream) can be seen in CFE-668.

pkg/controller/deployment/cert_manager_controller_set.go
```diff
@@ -24,7 +23,7 @@ type CertManagerControllerSet struct {
 func NewCertManagerControllerSet(
 	kubeClient kubernetes.Interface,
 	kubeInformersForTargetNamespace v1helpers.KubeInformersForNamespaces,
-	configClient configv1client.ConfigV1Interface,
+	// configClient configv1client.ConfigV1Interface,
```
swghosh, line 26: Nit: can we remove this line?

pkg/controller/deployment/generic_deployment_controller.go
```diff
@@ -64,7 +70,7 @@ func newGenericDeploymentController(
 			kubeInformersForTargetNamespace.Core().V1().Namespaces().Informer(),
 		},
 		controller,
-		openshiftClusterConfigClient,
+		&dummyClusterConfigClient{},
```
swghosh, line 73: IIUC, reading code from library-go seems like this config client is unused in terms of functionality within the controller, right?

[1] https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controller/workload/workload.go

swghosh: Correspondingly, then we could replace `&dummyClusterConfigClient{}` with `dummyClusterConfigClient{}`, if that works (by removing the additional pointer reference) without causing any type issues.

swghosh (commented):
Mostly LGTM, just left a minor comments, which are mostly nits / questions seeking some context to myself.
For the time being this looks like a harmless change given it helps remove ClusterOperator references in the cluster.

pkg/controller/deployment/generic_deployment_controller.go
thejasn: This gets completely removed in #85 [38c59a](https://github.com/openshift/cert-manager-operator/pull/85/commits/38c59a46e59f400ee1af924e58606b3590ad54bf).

thejasn:
/label docs-approved
Since this PR does not affect docs.

thejasn:
/cc @davemulford 
This fixes UI/UX issue https://issues.redhat.com/browse/CM-35.

thejasn:
Fixes CM-35

PR #87, opened 2022-12-15
[CFE-698] Unsupported config overrides to use union of default args

swghosh:
Currently, using UnsupportedConfigOverrides does not take the union of the default args. This change allows the override to work such that it can merge expected default values with user provided overrides.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/cert-manager/cert-manager v1.9.1
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/golangci/golangci-lint v1.50.1
+	github.com/google/go-cmp v0.5.9
 	github.com/google/go-jsonnet v0.17.0
 	github.com/mogensen/kubernetes-split-yaml v0.3.0
 	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
@@ -106,7 +107,6 @@ require (
 	github.com/golangci/revgrep v0.0.0-20220804021717-745bb2f7c2e6 // indirect
 	github.com/golangci/unconvert v0.0.0-20180507085042-28b1c447d1f4 // indirect
 	github.com/google/gnostic v0.6.9 // indirect
-	github.com/google/go-cmp v0.5.9 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 // indirect
 	github.com/google/uuid v1.3.0 // indirect
--- a/pkg/controller/deployment/unsupported_config_overrides.go
+++ b/pkg/controller/deployment/unsupported_config_overrides.go
@@ -1,23 +1,72 @@
 package deployment
 
 import (
+	"fmt"
+	"strings"
+
 	appsv1 "k8s.io/api/apps/v1"
 
 	"github.com/openshift/cert-manager-operator/api/operator/v1alpha1"
 )
 
+const argKeyValSeparator = "="
+
+// UnsupportedConfigOverrides overrides the values of container args in the deployment
 func UnsupportedConfigOverrides(deployment *appsv1.Deployment, unsupportedConfigOverrides *v1alpha1.UnsupportedConfigOverrides) *appsv1.Deployment {
 	if unsupportedConfigOverrides == nil {
 		return deployment
 	}
 	if len(unsupportedConfigOverrides.Webhook.Args) > 0 && deployment.Name == "cert-manager-webhook" {
-		deployment.Spec.Template.Spec.Containers[0].Args = unsupportedConfigOverrides.Webhook.Args
+		deployment.Spec.Template.Spec.Containers[0].Args = mergeContainerArgs(
+			deployment.Spec.Template.Spec.Containers[0].Args,
+			unsupportedConfigOverrides.Webhook.Args)
 	}
 	if len(unsupportedConfigOverrides.CAInjector.Args) > 0 && deployment.Name == "cert-manager-cainjector" {
-		deployment.Spec.Template.Spec.Containers[0].Args = unsupportedConfigOverrides.CAInjector.Args
+		deployment.Spec.Template.Spec.Containers[0].Args = mergeContainerArgs(
+			deployment.Spec.Template.Spec.Containers[0].Args,
+			unsupportedConfigOverrides.CAInjector.Args)
 	}
 	if len(unsupportedConfigOverrides.Controller.Args) > 0 && deployment.Name == "cert-manager" {
-		deployment.Spec.Template.Spec.Containers[0].Args = unsupportedConfigOverrides.Controller.Args
+		deployment.Spec.Template.Spec.Containers[0].Args = mergeContainerArgs(
+			deployment.Spec.Template.Spec.Containers[0].Args,
+			unsupportedConfigOverrides.Controller.Args)
 	}
 	return deployment
+}
+
+// mergeContainerArgs merges the source args with override values
+// using a map that tracks unique keys for each arg containing a
+// key value pair of form `key[=value]`
+func mergeContainerArgs(sourceArgs []string, overrideArgs []string) (destArgs []string) {
+	destArgMap := map[string]string{}
+	parseArgMap(destArgMap, sourceArgs)
+	parseArgMap(destArgMap, overrideArgs)
+
+	destArgs = make([]string, len(destArgMap))
+	i := 0
+	for key, val := range destArgMap {
+		if len(val) > 0 {
+			destArgs[i] = fmt.Sprintf("%s%s%s", key, argKeyValSeparator, val)
+		} else {
+			destArgs[i] = key
+		}
+		i++
+	}
+	return destArgs
+}
+
+// parseArgMap adds new entries to the map using keys
+// parsed from each arg (of the form `key=[value]`) from the
+// list of args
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/cert-manager/cert-manager v1.9.1
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/google/go-jsonnet v0.17.0
+	github.com/imdario/mergo v0.3.12
```
thejasn, line 9: I would prefer a solution without relying on external libraries. If logic https://github.com/openshift/library-go/blob/master/pkg/operator/unsupportedconfigoverridescontroller/unsupportedconfigoverrides_controller.go here doesn't solve our issue, then maybe we could look at external libraries. WDYT?

swghosh: As discussed, I've updated the implementation to not use mergo, as it isn't much of a help for the given use case. 
You could PTAL the latest changes.

swghosh:
Tested out the implemented changes on a cluster with the operator installed.

1. Before:
```
$ oc get deployments -n cert-manager -o yaml cert-manager
---
    spec:
      containers:
      - args:
        - --v=2
        - --cluster-resource-namespace=$(POD_NAMESPACE)
        - --leader-election-namespace=kube-system
        command:
        - /app/cmd/controller/controller
        env:
---
```

2. Changes - adding unsupported config overrides
```
$ oc edit certmanager cluster
apiVersion: operator.openshift.io/v1alpha1
kind: CertManager
metadata:
  creationTimestamp: "2022-12-19T13:35:27Z"
  generation: 4
  name: cluster
  resourceVersion: "5939223"
  uid: 0b8848ca-d4a7-4606-b1f9-6e4dfe1aaf57
spec:
  logLevel: Normal
  managementState: Managed
  observedConfig: null
  operatorLogLevel: Normal
  unsupportedConfigOverrides:
    controller:
      args:
      - --v=4
status:
---
```

3. After:
```
$ oc get deployments -n cert-manager -o yaml cert-manager
---
    spec:
      containers:
      - args:
        - --v=4
        - --cluster-resource-namespace=$(POD_NAMESPACE)
        - --leader-election-namespace=kube-system
        command:
        - /app/cmd/controller/controller
---
```
The arg for `--v` is getting overridden with the new changes (new values overridden for same key) which seems to have tackle the problem mentioned in #68.

pkg/controller/deployment/unsupported_config_overrides.go
```diff
@@ -1,23 +1,63 @@
...
+			destArgs[i] = key
+		}
+		i++
+	}
+	return destArgs
+}
+
+func updateArgMap(argMap map[string]string, args []string) {
+	for _, arg := range args {
+		splitted := strings.Split(arg, argKeyValSeparator)
```
thejasn, line 56: I say we skip on empty `arg` instead of adding empty entries into the map.

pkg/controller/deployment/unsupported_config_overrides.go
```diff
@@ -1,23 +1,63 @@
...
+			destArgs[i] = fmt.Sprintf("%s%s%s", key, argKeyValSeparator, val)
+		} else {
+			destArgs[i] = key
+		}
+		i++
+	}
+	return destArgs
+}
+
+func updateArgMap(argMap map[string]string, args []string) {
```
thejasn, line 54:
```suggestion
func parseArgMap(argMap map[string]string, args []string) {
```

thejasn: Also add a godoc on this function.

pkg/controller/deployment/unsupported_config_overrides_test.go
```diff
@@ -14,7 +14,12 @@ func TestUnsupportedConfigOverrides(t *testing.T) {
 	deploymentFile, err := assets.Asset("cert-manager-deployment/controller/cert-manager-deployment.yaml")
 	require.NoError(t, err)
 	testDeployment := resourceread.ReadDeploymentV1OrDie(deploymentFile)
-	defaultArgs := testDeployment.Spec.Template.Spec.Containers[0].Args
+
+	defaultArgs := make([]string, len(testDeployment.Spec.Template.Spec.Containers[0].Args))
+	copy(defaultArgs, testDeployment.Spec.Template.Spec.Containers[0].Args)
+
+	testArgs := []string{"--test-arg=true"}
```
thejasn, line 21: Can you also cover other variations of args?
```
--test=a
--test
--test=a=b=c
```

pkg/controller/deployment/unsupported_config_overrides_test.go
```diff
@@ -32,29 +37,31 @@ func TestUnsupportedConfigOverrides(t *testing.T) {
...
 		},
 		"Controller overrides should work": {
 			overrides: &v1alpha1.UnsupportedConfigOverrides{
 				Controller: v1alpha1.UnsupportedConfigOverridesForCertManagerController{
-					Args: []string{"test"},
+					Args: testArgs,
 				},
 			},
-			wantArgs: []string{"test"},
+			wantArgs: defaultPlusTestArgs,
```
thejasn, line 54: This only tests `append`  right?  because `testArgs` isn't a valid value in `deploymentFile`.

thejasn (changes requested):
Also could you change `TestUnsupportedConfigOverrides` -> `TestUnsupportedConfigOverridesForCertManagerController` or something along those lines since it doesn't test the other manifests anyways.

pkg/controller/deployment/unsupported_config_overrides_test.go
thejasn: You can cover this part in a separate unit test to cover `parseArgMap`/`updateArgMap`. Up to you.

pkg/controller/deployment/unsupported_config_overrides.go
```diff
@@ -1,23 +1,63 @@
...
 
+const argKeyValSeparator = "="
+
 func UnsupportedConfigOverrides(deployment *appsv1.Deployment, unsupportedConfigOverrides *v1alpha1.UnsupportedConfigOverrides) *appsv1.Deployment {
 	if unsupportedConfigOverrides == nil {
 		return deployment
 	}
 	if len(unsupportedConfigOverrides.Webhook.Args) > 0 && deployment.Name == "cert-manager-webhook" {
-		deployment.Spec.Template.Spec.Containers[0].Args = unsupportedConfigOverrides.Webhook.Args
+		deployment.Spec.Template.Spec.Containers[0].Args = overrideConfigArgs(
```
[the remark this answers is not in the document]
swghosh: Updated to use `mergeContainerArgs` as func name instead.

pkg/controller/deployment/unsupported_config_overrides.go
```diff
@@ -1,23 +1,63 @@
...
+			unsupportedConfigOverrides.Webhook.Args,
+			deployment.Spec.Template.Spec.Containers[0].Args)
 	}
 	if len(unsupportedConfigOverrides.CAInjector.Args) > 0 && deployment.Name == "cert-manager-cainjector" {
-		deployment.Spec.Template.Spec.Containers[0].Args = unsupportedConfigOverrides.CAInjector.Args
+		deployment.Spec.Template.Spec.Containers[0].Args = overrideConfigArgs(
+			unsupportedConfigOverrides.CAInjector.Args,
+			deployment.Spec.Template.Spec.Containers[0].Args)
```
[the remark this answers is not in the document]
swghosh: Yeah, good idea. Swapped the order of the arguments in func signature.

pkg/controller/deployment/unsupported_config_overrides.go
swghosh: Added a if block to guard around empty args like `if len(splitted) > 0 { ... }` which should be able to take care of that.

swghosh: Also, just to clarify in case you were referring to args containing only the key (`--key`) without any value as an empty arg, in that case updating the map with atleast the key would still be required otherwise we would end up missing such args defeating the purpose of the map.

pkg/controller/deployment/unsupported_config_overrides.go
swghosh: Thanks, added godoc(s).

pkg/controller/deployment/unsupported_config_overrides_test.go
swghosh: Added a test for `parseArgMap` with more extensive coverage (including similar cases as described).

pkg/controller/deployment/unsupported_config_overrides.go
thejasn: Hmm, what I meant by empty `arg` was if `args={"", "--foo", "--foo=bar"}`. Then again the `len()` would cover this case.

swghosh: Yeah, got it. I checked by putting `""` in the args list via `TestParseArgMap`, the len was still 1. Fixed it by updating the if statement to use: `if len(splitted) > 0 && arg != "" { ... }` instead, which helps achieve desired functionality now.

pkg/controller/deployment/unsupported_config_overrides_test.go
swghosh: Yes, that was correct. I added more test cases that covers the append as well as replace related behavior of UnsupportedConfigOverrides.

thejasn:
Fixes CM-23

PR #90, opened 2022-12-26
Fixes in ClusterServiceVersion manifest

swghosh:
Fix CSV manifests:
- trailing newline in `operators.openshift.io/valid-subscription` annotation which is required yaml lint
- description in base had erroneous `'displayName:..'` which was removed
- populated config/samples with examples of each resource type
    - during bundle phase it adds the `alm-examples` annotation to CSV
    - which helps avoid `WARN[0000] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation`


config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   annotations:
-    alm-examples: '[]'
```
thejasn, line 5: You can still keep this. It will get over written.

swghosh: Okay, done 👍🏻

thejasn:
/label docs-approved
/label px-approved
Only minor bundle changes, no impact on docs/px.

swghosh:
/cc @xingxingxia 
for qe-approved label

PR #91, opened 2022-12-30
Fix namespace in Makefile, change to default namespace in alm-examples

swghosh:
As a part of #83 we changed the operator namespace to use "cert-manager-operator" instead of "openshift-cert-manager-operator". In certain files such as README.md (upgrade section) as well as Makefile (local-run target) the value of the namespace reflects old value. This is intended at changing those values.

Additionally, the entries in alm-examples of the ClusterServiceVersion used to have `namespace: sample-namespace`, this PR would change it to `namespace: default` easing in applying the manifests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ verify-deps:
 	hack/verify-deps.sh
 
 local-run: build
-	./cert-manager-operator start --config=./hack/local-run-config.yaml --kubeconfig=$${KUBECONFIG:-$$HOME/.kube/config} --namespace=openshift-cert-manager-operator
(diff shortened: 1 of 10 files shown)


swghosh:
/cc @ayesha54 
replaced the namespace in each entry of alm-examples to use "default" instead of "sample-namespace" which could ease applying those manifests without the need to create an extra new namespace

swghosh:
/cc @thejasn @TrilokGeer 
PTAL

xingxingxia:
LGTM
/label qe-approved

swghosh:
/label px-approved
/label docs-approved
This PR isn't intended at any significant user-facing changes.

PR #93, opened 2023-01-16
CFE-740: Updated cert-manager-operator API to enable customization

thejasn:
## Description

Update operator API to support customization of operand deployments (controller, webhook, cainjector). This reduces end user dependency on using `UnsupportedConfigOverrides` which is not supported as the name indicates.


bundle/manifests/operator.openshift.io_certmanagers.yaml
```diff
@@ -34,6 +34,263 @@ spec:
...
+                      type: string
+                    type: array
+                  overrideEnv:
+                    items:
+                      description: EnvVar represents an environment variable present
+                        in a Container.
+                      properties:
+                        name:
+                          description: Name of the environment variable. Must be a
+                            C_IDENTIFIER.
```
TrilokGeer, line 57: Is is possible to have regular expression pattern?

thejasn: Yes, kubebuilder does provide regex pattern. `+kubebuilder:validation:Pattern:
string` . ([ref](https://book.kubebuilder.io/reference/markers/crd-validation.html))

config/crd/bases/operator.openshift.io_certmanagers.yaml
```diff
@@ -35,6 +35,263 @@ spec:
           spec:
             description: CertManagerSpec defines the desired state of CertManager
             properties:
+              cainjectorConfig:
+                description: "CAInjectorConfig specifies further customization options
+                  for the cainjector's deployment spec. \n Possible customizations
+                  include the following, \n For OverrideArgs, --namespace string                          If
+                  set, this limits the scope of cainjector to a single namespace.
+                  If set, cainjector will not update resources with certificates outside
+                  of the configured namespace. --v Level                                   number
```
TrilokGeer, line 44: nit :extra spaces

thejasn:
/cc @xingxingxia 
This will be followed up with #95, so most of the testing would need to be done as part of that.

xingxingxia:
Intended to review/research to do a pre-merge test before labeling. Due to no enough time currently, let me first add it:
/label qe-approved
Will test after merge.

PR #95, opened 2023-01-24
Feature/extend operator API

thejasn:


pkg/controller/deployment/unsupported_config_overrides.go
```diff
@@ -1,80 +0,0 @@
-package deployment
```
TrilokGeer, line 1: Any significance for changing file name? Same filename helps to provide function diffs.

pkg/operator/starter.go
```diff
@@ -62,7 +62,9 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		kubeClient,
 		kubeInformersForTargetNamespace,
 		kubeInformersForTargetNamespace.InformersFor(operatorclient.TargetNamespace),
-		operatorClient, resourceapply.NewKubeClientHolder(kubeClient).WithAPIExtensionsClient(apiExtensionsClient),
+		operatorClient,
+		certManagerInformers,
```
TrilokGeer, line 66: nit : It is a good practice to add new vars at the last of the list to avoid any conflicts arising when there are multiple contributions on the same. for now, the suggestion can be ignored.

pkg/controller/deployment/unsupported_config_overrides.go
thejasn: This PR adds a few more files/hooks that work on deployments. `deployment_validation*`, `deployment_overrides*` and their corresponding tests. Wanted to keep all the hooks in their respective files. Since `unsupported_config_overrides` was also changed to be hook func, renaming the file to follow the same naming pattern of the hooks made sense.

thejasn:
Fixes CM-11

PR #97, opened 2023-01-25
Fix ordered args override

thejasn:
### Description
Un-ordered override causes deployment hash to be different causing repetitive deployment updates. This fixes the order only when container args are overridden.

### DIFF
--- a/pkg/controller/deployment/unsupported_config_overrides.go
+++ b/pkg/controller/deployment/unsupported_config_overrides.go
@@ -2,6 +2,7 @@ package deployment
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


swghosh:
/label docs-approved
/label px-approved
This PR does not introduce any user-facing changes.

swghosh (approved):

swghosh:
/label qe-approved
Adding the label myself as the QE guys are out (unavailable) this week. Also, this shouldn't bring any changes to user test cases.

thejasn:
/cc @xingxingxia 
PTAL once you are available.

PR #101, opened 2023-02-02
CFE-754: Add seccompProfile RuntimeDefault to operator deployment

swghosh:
Add seccompProfile RuntimeDefault to deployment which helps avoid pod security violation of operator pod(s).

### DIFF
--- a/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
@@ -653,6 +653,8 @@ spec:
                     type: RuntimeDefault
               securityContext:
                 runAsNonRoot: true
+                seccompProfile:
+                  type: RuntimeDefault
               serviceAccountName: cert-manager-operator-controller-manager
               terminationGracePeriodSeconds: 10
       permissions:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -36,6 +36,8 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - command:
           - /usr/bin/cert-manager-operator


swghosh:
Operator deployed with these changes applied gets correct scc annotations.

```diff
$ oc get pods -n cert-manager-operator cert-manager-operator-controller-manager-7f65dc5697-wbx4g -o json | jq '.metadata'
{
  "annotations": {
    "k8s.ovn.org/pod-networks": "{\"default\":{\"ip_addresses\":[\"10.128.2.13/23\"],\"mac_address\":\"0a:58:0a:80:02:0d\",\"gateway_ips\":[\"10.128.2.1\"],\"ip_address\":\"10.128.2.13/23\",\"gateway_ip\":\"10.128.2.1\"}}",
    "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.13\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:0d\",\n    \"default\": true,\n    \"dns\": {}\n}]",
    "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.13\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:0d\",\n    \"default\": true,\n    \"dns\": {}\n}]",
    "kubectl.kubernetes.io/default-container": "cert-manager-operator",
+    "openshift.io/scc": "restricted-v2",
+   "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
  },
  "creationTimestamp": "2023-02-02T07:28:25Z",
  "generateName": "cert-manager-operator-controller-manager-7f65dc5697-",
  "labels": {
    "name": "cert-manager-operator",
    "pod-template-hash": "7f65dc5697"
  },
  "name": "cert-manager-operator-controller-manager-7f65dc5697-wbx4g",
  "namespace": "cert-manager-operator",
  "ownerReferences": [
    {
      "apiVersion": "apps/v1",
      "blockOwnerDeletion": true,
      "controller": true,
      "kind": "ReplicaSet",
      "name": "cert-manager-operator-controller-manager-7f65dc5697",
      "uid": "f9cb2208-305d-4591-ae7d-b418a1ce3718"
    }
  ],
  "resourceVersion": "31922",
  "uid": "51067b80-329d-489d-ba9b-08dff8b6be0d"
}
```

swghosh:
I tried out running the operator pod with current changes from master, and I can now confirm that without these changes the pods get the right annotations too.

```diff
$ oc get pods -n cert-manager-operator cert-manager-operator-controller-manager-7877b5bf49-929mk -o json | jq '.metadata'
{
  "annotations": {
    "k8s.ovn.org/pod-networks": "{\"default\":{\"ip_addresses\":[\"10.128.2.22/23\"],\"mac_address\":\"0a:58:0a:80:02:16\",\"gateway_ips\":[\"10.128.2.1\"],\"ip_address\":\"10.128.2.22/23\",\"gateway_ip\":\"10.128.2.1\"}}",
    "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.22\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:16\",\n    \"default\": true,\n    \"dns\": {}\n}]",
    "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.22\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:16\",\n    \"default\": true,\n    \"dns\": {}\n}]",
    "kubectl.kubernetes.io/default-container": "cert-manager-operator",
+    "openshift.io/scc": "restricted-v2",
+    "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
  },
  "creationTimestamp": "2023-02-02T08:58:16Z",
  "generateName": "cert-manager-operator-controller-manager-7877b5bf49-",
  "labels": {
    "name": "cert-manager-operator",
    "pod-template-hash": "7877b5bf49"
  },
  "name": "cert-manager-operator-controller-manager-7877b5bf49-929mk",
  "namespace": "cert-manager-operator",
  "ownerReferences": [
    {
      "apiVersion": "apps/v1",
      "blockOwnerDeletion": true,
      "controller": true,
      "kind": "ReplicaSet",
      "name": "cert-manager-operator-controller-manager-7877b5bf49",
      "uid": "c9e715fc-4278-4c89-a438-54e46f6db2f9"
    }
  ],
  "resourceVersion": "64525",
  "uid": "6d4a71f1-6764-4b7a-a071-502043c9453b"
}
```

But as a nit would still prefer pushing for this change, as currently the container security context have the right seccomp profiles but it'd be good for ensuring that the pod security context has it added to, for completeness.

/cc @thejasn

thejasn:
> Operator deployed with these changes applied gets correct scc annotations.
> 
> ```diff
> $ oc get pods -n cert-manager-operator cert-manager-operator-controller-manager-7f65dc5697-wbx4g -o json | jq '.metadata'
> {
>   "annotations": {
>     "k8s.ovn.org/pod-networks": "{\"default\":{\"ip_addresses\":[\"10.128.2.13/23\"],\"mac_address\":\"0a:58:0a:80:02:0d\",\"gateway_ips\":[\"10.128.2.1\"],\"ip_address\":\"10.128.2.13/23\",\"gateway_ip\":\"10.128.2.1\"}}",
>     "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.13\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:0d\",\n    \"default\": true,\n    \"dns\": {}\n}]",
>     "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"ovn-kubernetes\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.128.2.13\"\n    ],\n    \"mac\": \"0a:58:0a:80:02:0d\",\n    \"default\": true,\n    \"dns\": {}\n}]",
>     "kubectl.kubernetes.io/default-container": "cert-manager-operator",
> +    "openshift.io/scc": "restricted-v2",
> +   "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
>   },
>   "creationTimestamp": "2023-02-02T07:28:25Z",
>   "generateName": "cert-manager-operator-controller-manager-7f65dc5697-",
>   "labels": {
>     "name": "cert-manager-operator",
>     "pod-template-hash": "7f65dc5697"
>   },
>   "name": "cert-manager-operator-controller-manager-7f65dc5697-wbx4g",
>   "namespace": "cert-manager-operator",
>   "ownerReferences": [
>     {
>       "apiVersion": "apps/v1",
>       "blockOwnerDeletion": true,
>       "controller": true,
>       "kind": "ReplicaSet",
>       "name": "cert-manager-operator-controller-manager-7f65dc5697",
>       "uid": "f9cb2208-305d-4591-ae7d-b418a1ce3718"
>     }
>   ],
>   "resourceVersion": "31922",
>   "uid": "51067b80-329d-489d-ba9b-08dff8b6be0d"
> }
> ```

Since containers have the correct `seccompProfile` set it doesn't raise an issue. As for as I know pod security context is applied first followed by container security context which overrides the values. I know for fields like `runAsUser` container security context overrides pod security context but I'm not sure if that is case for `seccompProfile` as well. If that is the case, then adding to pod security context behaves similar to a default.

swghosh:
>Since containers have the correct seccompProfile set it doesn't raise an issue. As for as I know pod security context is applied first followed by container security context which overrides the values. I know for fields like runAsUser container security context overrides pod security context but I'm not sure if that is case for seccompProfile as well. If that is the case, then adding to pod security context behaves similar to a default.


This is some of the code I found in upstream k8s: https://github.dev/kubernetes/kubernetes/blob/release-1.26/pkg/api/pod/warnings.go#L210-L215, regarding raising warnings for pod's seccompProfile, but it shouldn't exactly related to our current context. 

However, I did find OpenShift adding the `seccompProfile: type: "RuntimeDefault"` to the deployment similar to what we are doing now, so ultimately no warning as container security context is seccomp compliant; so upto us if we want to add it in pod securitycontext additionally or not.

IMHO, now the only possible perk would be to use the same RuntimeDefault seccompProfile in future side containers we add to the same deployment where we do not add their own container security context, no other advantage apart from that in this case I guess.

xingxingxia:
Checked your comments about the reason you keep it. It is OK to me. About the question, yes it overrides: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context-2 :
General:
SecurityContext defines the security options the ephemeral container should be run with. If set, the fields of SecurityContext override the equivalent fields of PodSecurityContext.

Specific:
If seccomp options are provided at both the pod & container level, the container options override the pod options.

/label qe-approved

swghosh:
/cc @bergerhoffer @davemulford
for docs-approved and px-approved labels respectively

PR #102, opened 2023-02-02
CFE-694: CSV changes for olm release channel, version, annotations

swghosh:
CSV changes:
- [x] **Annotations**: 
    - [x] alm-examples
    - [x] ~~operatorframework.io/initialization-resource~~
    - [x] operatorframework.io/suggested-namespace
    - [x] ~~operators.openshift.io/infrastructure-features~~
    - [x] operators.openshift.io/valid-subscription
    - [x] ~~operators.operatorframework.io/internal-objects~~
- [x] **Channels**
    - [x] stable-v1, stable-v1.10 
- [x] **Bundle Version**: 1.10.2
- [x] **More annotations**:
    - [x] **installMode**
    - [x] createdAt
    - [x] containerImage


swghosh:
Referring to the doc on OpenShift specific annotations (https://docs.openshift.com/container-platform/4.12/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-manual-annotations_osdk-generating-csvs) on the operator CSV, missing out on the following currently:
1. `operatorframework.io/initialization-resource`: not needed as of now, can add in future if needed to add a specific resource to the cluster upon operator installation.
2. `operators.openshift.io/infrastructure-features`: have added `['disconnected', 'fips']` for now, can verify if proxy-aware is supported and append that if required.
3. `operators.operatorframework.io/internal-objects`: not needed as of now, all CRDs installed by this operator are ought to be shown in OpenShift console.

swghosh:
As per our last discussion, we'll not be using `operators.openshift.io/infrastructure-features` or `operators.operatorframework.io/internal-objects` as of now.

swghosh:
> The entry specifying the skipRange will be presented as a direct (one hop) update to any version from that package within that range. The versions in this range do not need to be in the catalog in order for bundle addition to be successful. We recommend avoiding using unbound ranges such as <1.0.3.

ref: https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange

> When searching catalogs for updates, if the head of a channel has a skipRange annotation and the currently installed Operator has a version field that falls in the range, OLM updates to the latest entry in the channel.

ref: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.2/html/operators/understanding-the-operator-lifecycle-manager-olm > 2.1.3.3. Replacing multiple Operators

Hence, not using skipRange for the first release as we don't want to upgrade this GA release from any existing TechPreview releases, the operator is ought to be installed newly onto a cluster. However, we should ensure to add a suitable value to `olm.skipRange` when releasing this operator next time after first release.

swghosh:
/cc @bergerhoffer for docs-approved label
/cc @davemulford for px-approved label
/cc @xingxingxia for qe-approved label
/cc @thejasn for reviews/approval

Thanks!

PR #105, opened 2023-02-17
Use OpenShift subscription as 'OKE and up'

swghosh:
Change CSV annotation `operators.openshift.io/valid-subscription` to support 'OKE and up' subscriptions instead of current 'OCP and up' as required by PM.


davemulford:
LGTM
/label px-approved

swghosh:
/cc @xingxingxia 
xref: https://redhat-internal.slack.com/archives/C045YMPKR3M/p1676651990958149?thread_ts=1676480756.122199&cid=C045YMPKR3M

PR #111, opened 2023-03-21
OCPBUGS-9994: Add logLevel hook for setting --v of operands

swghosh:
New deployment hook for adding `--v` argument on the basis of logLevel in operatorSpec.


swghosh:
/hold
I'll check out another operator like openshift-apiserver-operator to check if the values for --v is identical for Trace and TraceAll levels.

thejasn:
The bug also mentions `operatorLogLevel`, are you planning on addressing that in this PR itself?

pkg/controller/deployment/operator_log_level.go
```diff
@@ -0,0 +1,37 @@
...
+
+// setOperatorLogLevel is used to set the value of klog verbosity
+// if current verbosity level does not match logLevel
+func setOperatorLogLevel(logLevel operatorv1.LogLevel) error {
+	if logLevel == "" {
+		logLevel = operatorv1.Normal
+	}
+
+	v := klog.Level(operatorLogLevels[logLevel])
+	// set verbosity by checking if V(n) and V(n+1) are enabled,
```
thejasn, line 27: Can you add a comment on why `V(n)` and `V(n+1)` are considered as well?

pkg/controller/deployment/operator_log_level.go
```diff
@@ -0,0 +1,37 @@
...
+	// change runtime verbosity level iff !V(n) or [V(n) and V(n+1)]
+	if !klog.V(v).Enabled() || (klog.V(v).Enabled() && klog.V(v+1).Enabled()) {
+		msg, err := logs.GlogSetter(strconv.Itoa(int(v)))
+		if err != nil {
+			return err
+		}
```
thejasn, line 33: Can this possible result in a non-nil error at all?

pkg/controller/deployment/operator_log_level_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func TestSetOperatorLogLevel(t *testing.T) {
+	testCases := []struct {
+		logLevel              operatorv1.LogLevel
+		desiredVerbosityLevel int
+	}{
+		{
+			logLevel:              operatorv1.Normal,
+			desiredVerbosityLevel: 2,
```
thejasn, line 18: Can you add negative cases as well?

pkg/controller/deployment/operator_log_level.go
swghosh: I would say the chances are low but always good to have a defensive code path since in `logs.GlogSetter` they do already have an error case added:
https://github.com/openshift/cert-manager-operator/blob/ocpbugs-9994/vendor/k8s.io/component-base/logs/logs.go#L185-L193

pkg/controller/deployment/operator_log_level.go
swghosh: Yeah, I can do that; context being there isn't a direct straight-forward way to check the current verbosity level at runtime (atleast that I know of going through `klog` and `k8s.io/component-base/logs`) which brought me up to this logic to compare with the help of V(n) and V(n+1) enabled or not at runtime.

thejasn (commented):
Can you add sample logs of before and after for both the operand and operator? Based on this I want to decide if we want to support `operatorLogLevel` (via the operator spec) at all.

swghosh:
> Since the operator already has
> 
> https://github.com/openshift/cert-manager-operator/blob/70e39c50c56a5819e05b46b6d34be53d7fef0951/main.go#L23
> , Is this functional? If yes we should be able to control log level via env variables which can be updated via the subscription.
> 
> Assuming the above flow works, I think we can choose not to support the operator `logLevel` via the `cluster` scoped resource.

As discussed previously, to support updating the operator log level from the deployment spec (possibly via the Subscription object on the cluster under the most likely use case of operator being installed through console) and given the predicate that we indeed want to do away with the default controller in the near future, I've updated the PR accordingly with ensuring that `operatorLogLevel` field of our custom resource is not used for setting the verbosity. 

Despite the `pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)` I found out that our binary still lacked the `--v` flag, and I was able to add it with `cli.Run(command)` referring to [how cluster-kube-api-server-operator adds those klog verbosity flags](https://github.com/openshift/cluster-kube-apiserver-operator/blob/ee6d396314325597fbfd32bfb5d05824e0d51d06/cmd/cluster-kube-apiserver-operator/main.go#L31).

pkg/controller/deployment/operator_log_level.go
swghosh: Updates in conversation thread beneath. This part of the implementation have been thoroughly changed in new commits.

pkg/controller/deployment/operator_log_level.go
swghosh: https://github.com/openshift/cert-manager-operator/pull/111#discussion_r1169705175

pkg/controller/deployment/operator_log_level_test.go
swghosh: https://github.com/openshift/cert-manager-operator/pull/111#discussion_r1169705175

swghosh:
Changed cli flags summary:
```diff
$ ./cert-manager-operator --help
OpenShift cluster cert-manager operator

Usage:
  cert-manager-operator [flags]
  cert-manager-operator [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  start       Start the cert-manager Operator

Flags:
   -h, --help                           help for cert-manager-operator
       --kubeconfig string              Paths to a kubeconfig. Only required if out-of-cluster.
+      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)
+  -v, --v Level                        number for the log level verbosity
+      --vmodule moduleSpec             comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log format)

Use "cert-manager-operator [command] --help" for more information about a command.

```

swghosh:
Circling back to our discussion on allowing users to choose to change the operator log level by modifying the subscription object, I've added the `OPERATOR_LOG_LEVEL` env variable and set it value in the `--v` arg for klog to setup the log level correctly. 

(JFYI @thejasn I found out that [node-observability-operator uses a similar setting](https://github.com/openshift/node-observability-operator/blob/40aaadae753f6bf385c9a171933c75faa8723c8b/config/manager/manager.yaml#L34-L39) but since it is based on controller runtime it leverages zap logger which has a different log levels compared to klog, to my understanding [debug in zaplog is probably the highest level available](https://github.com/uber-go/zap/blob/845ca51d5b8d9fed9fe14f35ab13b6b160d5762d/level.go#L30-L49) even using which we don't get the higher level klog logs that Trace (--v=6) and TraceAll (--v=8) gives us from most of library-go code, so added it this way).

I'll add a md doc about updating operator log level on production cluster (via subscription object) as well as while testing either through `--v` or changing the deployment's env var.

ref: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md

thejasn:
/remove-jira/invalid-bug

xingxingxia:
This hasn't yet Dev /lgtm label before labeling qe-approved. But anyway, labeling it; would like to do post-merge test since other work (including OCP Auth) items need me :)
/label qe-approved

swghosh:
/cc @xenolinux 
for docs-approved label
/cc @davemulford 
for px-approved label

PR #112, opened 2023-04-03
OCPBUGS-8466: Add operand pod label customization

swghosh:
- API add overrideLabels field to DeploymentConfig of cert-manager controller, cainjector, webhook operands
- custom labels to be be added to through deployment hook


thejasn (changes requested):
Since the logic present here merges the existing labels with the new provided labels, I think we want to some validation before merging. Possibly only allowing labels from `certmanager.io`?

pkg/controller/deployment/deployment_helper.go
```diff
@@ -76,6 +76,17 @@ func parseArgMap(argMap map[string]string, args []string) {
 	}
 }
 
+func mergeLabels(sourceLabels map[string]string, overrideLabels map[string]string) map[string]string {
```
thejasn, line 79: Can you also add a few unit tests for this func as well (similar to the other merge funcs)?

thejasn: Even better, you can re-use https://github.com/kubernetes/apimachinery/blob/31e4b9d241f7bb8490585cc57a0e5ee5a068b432/pkg/labels/labels.go#L118

thejasn: The function signature will change everywhere to `labels.Set` but should be acceptable since there are specific functions for labels.

swghosh: Yup, updated to use `labels.Merge` function in the implementation.

swghosh:
> Since the logic present here merges the existing labels with the new provided labels, I think we want to some validation before merging.

Added a validation that filters out conflicting labels present in `deployment.Spec.Selector.MatchLabels` to avoid deployment pods being mismatched.

swghosh:
> Since the logic present here merges the existing labels with the new provided labels, I think we want to some validation before merging. Possibly only allowing labels from certmanager.io?

@thejasn 
added the `withPodLabelsValidateHook()` to ensure we validate and allow only podLabels that should be supported. For now, I've added `azure.workload.identity/use` label (xref: https://cert-manager.io/docs/configuration/acme/dns01/azuredns/#reconfigure-cert-manager, for Azure Workload Identity support) and we can update the list with other labels on need basis as discussed previously.

xingxingxia:
/label qe-approved
Would like to do post-merge test : ) since other work (including OCP Auth) items need me :)

PR #113, opened 2023-04-06
[WIP] Rebase cert-manager v1.11

swghosh:


go.mod
```diff
@@ -3,36 +3,37 @@ module github.com/openshift/cert-manager-operator
 go 1.19
 
 require (
-	github.com/cert-manager/cert-manager v1.10.2
+	github.com/cert-manager/cert-manager v1.11.0
```
thejasn, line 6: I think we should be consuming our downstream fork here, right?

go.mod
```diff
@@ -3,36 +3,37 @@ module github.com/openshift/cert-manager-operator
...
 	github.com/golangci/golangci-lint v1.50.1
 	github.com/google/go-cmp v0.5.9
 	github.com/google/go-jsonnet v0.17.0
 	github.com/mogensen/kubernetes-split-yaml v0.3.0
-	github.com/openshift/api v0.0.0-20230120195050-6ba31fa438f2
+	github.com/openshift/api v0.0.0-20230330150608-05635858d40f
 	github.com/openshift/build-machinery-go v0.0.0-20220913142420-e25cf57ea46d
 	github.com/openshift/client-go v0.0.0-20230120202327-72f107311084
-	github.com/openshift/library-go v0.0.0-20230120214501-9bc305884fcb
+	github.com/openshift/library-go v0.0.0-20230405224819-c1e9763926fb
```
thejasn, line 15: Were these to updated explicitly or bumped as a dependency?

go.mod
swghosh: I took a note of this during last rebase, but probably left it untouched as we've been doing it this way since before. Assuming we push the exact same commits to our fork does it really matter? However, if you suggest adding it we can, probably just a matter of another `replace(...)` block you're mentioning of right?

go.mod
swghosh: I bumped explicitly. Not sure, if it is absolutely required but better to stay updated with latest and greatest (just hoping that 4.12, 4.13 APIs don't give us any problems)!

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -733,4 +733,4 @@ spec:
     name: cert-manager-ca-injector
   - image: quay.io/jetstack/cert-manager-controller:v1.10.2
```
thejasn, line 734: The related images need to updated as well.

pkg/controller/deployment/deployment_overrides_test.go
```diff
@@ -92,9 +92,7 @@ func TestUnsupportedConfigOverrides(t *testing.T) {
 					Args: testArgsToAppend,
 				},
 			},
-			wantArgs: []string{
-				"--v=2", "--cluster-resource-namespace=$(POD_NAMESPACE)", "--leader-election-namespace=kube-system",
-			},
+			wantArgs: deploymentDefaultArgs["cert-manager"],
```
thejasn, line 95: Tests shouldn't be re-using the code that it testing. So I would prefer to have the explicit strings here instead of using the variable that is contains values from the manifest file (we want to test what is in the manifest file).

thejasn: I would prefer if this was the case in the other tests too, but lets not actively revert this.

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
swghosh: Updated.

pkg/controller/deployment/deployment_overrides_test.go
swghosh: Thanks, I've made these changes in: 3f0b8ea0a2ca370d4121c9b54b2d0a16aa3f7bb8 to ensure the said behaviour is followed.

thejasn:
@swghosh Can you update the description to contains all the `go.mod` changes being done along with the new versions. (library-go, controller-runtime, k8s deps, etc that were manually bumped)

swghosh:
@thejasn I've updated the PR description with all the relevant information.

xingxingxia:
Adding qe-approved based on pre-merge test results in channel https://redhat-internal.slack.com/archives/C045YMPKR3M/p1682073637235019?thread_ts=1681844002.801959&cid=C045YMPKR3M 
/label qe-approved

thejasn:
cc: @davemulford for px-approved
cc: @xenolinux for docs-approved

davemulford:
I don't see anything major in the upstream docs about upgrading to this version.
/label px-approved

PR #116, opened 2023-04-24
Fix proxy.md

xingxingxia:
/cc @swghosh @thejasn 
The display in https://github.com/openshift/cert-manager-operator/blob/master/docs/proxy.md does not hide ```bash, it displays as:

        ```bash
        oc -n cert-manager-operator patch subscription cert-manager-operator ...
        ```

Also, the subscription name (if installed via web) is `openshift-cert-manager-operator` instead. https://docs.openshift.com/container-platform/4.12/security/cert_manager_operator/cert-manager-operator-proxy.html needs update to be `subscription openshift-cert-manager-operator` as well (didn't catch this when GA sorry), CC @bergerhoffer .

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -18,9 +18,9 @@ Follow the instructions below to let Cert Manager Operator trust a custom Certif
 
 2.  Consume the created configmap in Cert Manager Operator's deployment by updating its subscription:
 
-        ```bash
-        oc -n cert-manager-operator patch subscription cert-manager-operator --type='merge' -p '{"spec":{"config":{"env":[{"name":"TRUSTED_CA_CONFIGMAP_NAME","value":"trusted-ca"}]}}}'
-        ```
+    ```bash
+    oc -n cert-manager-operator patch subscription <subscription_name> --type='merge' -p '{"spec":{"config":{"env":[{"name":"TRUSTED_CA_CONFIGMAP_NAME","value":"trusted-ca"}]}}}'
+    ```
 
     _Note_: Alternatively, you can also patch the `cert-manager-operator-controller-manager` deployment in the `cert-manager-operator` namespace.
     `bash


thejasn:
cc: @xenolinux

swghosh:
/assign @xenolinux 
if you could review, it'd be great as we need to update the docs accordingly

swghosh:
/label px-approved
/label qe-approved
No functional changes are being introduced by this change.

docs/proxy.md
```diff
@@ -18,9 +18,9 @@ Follow the instructions below to let Cert Manager Operator trust a custom Certif
 
 2.  Consume the created configmap in Cert Manager Operator's deployment by updating its subscription:
 
-        ```bash
-        oc -n cert-manager-operator patch subscription cert-manager-operator --type='merge' -p '{"spec":{"config":{"env":[{"name":"TRUSTED_CA_CONFIGMAP_NAME","value":"trusted-ca"}]}}}'
-        ```
+    ```bash
+    oc -n cert-manager-operator patch subscription openshift-cert-manager-operator --type='merge' -p '{"spec":{"config":{"env":[{"name":"TRUSTED_CA_CONFIGMAP_NAME","value":"trusted-ca"}]}}}'
```
swghosh, line 22: @xingxingxia after giving this one a re-thought, IMO can we please keep this one as is in case of this GitHub repo and just update this to use "openshift-cert-manager-operator" only in https://github.com/openshift/openshift-docs 
The rationale behind it being: users installing the operator from GitHub source will continue to use subscription name as "cert-manager-operator" while those who installed the productized RH version of the operator from OperatorHub (rh operator-index) would have subscription object named "openshift-cert-manager-operator" (which is becuase we patch the operator package id in midstream before building the productized version).

xingxingxia: Sure, but I updated with placeholder "<subscription_name>"

swghosh: Yeah, that's a good idea :)

swghosh:
/label docs-approved
PR for doc change is up: https://github.com/openshift/openshift-docs/pull/61748

PR #118, opened 2023-04-25
OCPBUGS-10821: Update ACME solver image to read from env 

thejasn:
Description
---

Adds a new ENV variable to read the ACME solver image at runtime. Related PR on midstream will update the image to reference Red Hat's image. 

Related Midstream PR: https://gitlab.cee.redhat.com/cpaas-midstream-cfe/openshift-cert-manager/-/merge_requests/96

PR #119, opened 2023-04-26
Add verify-bundle step to CI

thejasn:
/hold until #118


thejasn:
/label px-approved
/label docs-approved
/label qe-approved 

This change only affects dev workflow. No impact on other areas.

PR #120, opened 2023-04-26
OCPBUGS-9998: Ambient credentials support

swghosh:
- Add `--issuer-ambient-credentials` as supported cert-manager controller arg
- Service account bound token using projected volume "/var/run/secrets/openshift/serviceaccount/token"


swghosh:
/hold
till we figure out if mounting the cloud credentials secret to the deployment pod itself is required or just bound-sa-token would suffice

swghosh:
/hold
the secrets informer implemented in this PR has an empty cache (probably, a bug or something we're doing wrong during initialization)

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,141 @@
...
+				},
+			},
+		}}
+		volumeMounts := []corev1.VolumeMount{{
+			Name:      boundSATokenVolumeName,
+			MountPath: boundSATokenDir,
+			ReadOnly:  true,
+		}}
+
+		if secretName := os.Getenv("CLOUD_SECRET_NAME"); secretName != "" {
```
stlaz, line 96: don't you have operator configuration that you can read this from?

stlaz, line 96:
```go
secretName := os.Getenv("CLOUD_SECRET_NAME")
if len(secretName) == 0 {
    return nil
}
```

stlaz (changes requested):

pkg/controller/deployment/generic_deployment_controller.go
```diff
@@ -48,5 +50,6 @@ func newGenericDeploymentController(
 		withUnsupportedArgsOverrideHook,
 		withProxyEnv,
 		withCAConfigMap(kubeInformersForTargetNamespace.Core().V1().ConfigMaps(), deployment, trustedCAConfigmapName),
+		withCloudCredentials(kubeInformersForTargetNamespace.Core().V1().Secrets(), configInformer.Config().V1().Infrastructures(), deployment.Name),
```
stlaz, line 53: Add the secrets informer to the informer array on line 39.

Not only it will add a wait for its caches to sync upon the controller startup, it's also going to add event handlers to the informer events. Without it, your controller is not going to react to the secret changes.

stlaz (changes requested):

pkg/controller/deployment/credentials_request.go
swghosh: yup, had this one improvised in another follow-up commit, yet to push it.

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,147 @@
...
+								Path: cloudCredentialConfig.mountFilename,
+							}},
+						},
+					},
+				})
+				volumeMounts = append(volumeMounts, corev1.VolumeMount{
+					Name:      cloudCredentialsVolumeName,
+					MountPath: cloudCredentialConfig.mountDirectory,
+				})
+			}
```
thejasn, line 133: Can you change this to a switch and drop `cloudCredentialConfigs`? I think it would make it more readable and maintainable.

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,147 @@
...
+						}},
+					},
+				},
+			},
+		}}
+		volumeMounts := []corev1.VolumeMount{{
+			Name:      boundSATokenVolumeName,
+			MountPath: boundSATokenDir,
+			ReadOnly:  true,
+		}}
```
thejasn, line 95: Can you move this to a separate hook? Since these are more like defaults for an operator running on OCP.

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,147 @@
...
+			_, err := secretsInformer.Lister().Secrets(operatorclient.TargetNamespace).Get(secretName)
+			if err != nil {
+				return err
+			}
+
+			if err != nil && apierrors.IsNotFound(err) {
+				return fmt.Errorf("(Retrying) cloud secret %q doesn't exist due to %v", secretName, err)
+			} else if err != nil {
+				return err
+			}
```
thejasn, line 107: Duplicate error handling. Merge them.

pkg/controller/deployment/deployment_overrides_validation.go
```diff
@@ -32,6 +32,13 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
 		"--v", "-V",
 		// The host and port that the metrics endpoint should listen on. (default "0.0.0.0:9402")
 		"--metrics-listen-address",
+		// Whether an issuer may make use of ambient credentials.
+		// 'Ambient Credentials' are credentials drawn from the environment, metadata services,
+		// or local files which are not explicitly configured in the Issuer API object.
+		// When this flag is enabled, the following sources for credentials are also used:
+		// AWS - All sources the Go SDK defaults to,
+		// notably including any EC2 IAM roles available via instance metadata.
```
thejasn, line 40: Add note on specific requirement for GCP.

pkg/operator/starter.go
```diff
@@ -58,20 +64,28 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
 		"",
 		"kube-system",
-		"cert-manager",
```
thejasn, line 61: We want this informer as well, to enforce resources created in `cert-manager` namespace.

thejasn (changes requested):
Can you update docs as well to include required commands on how to set credentials secret name?

pkg/operator/starter.go
swghosh: `operatorclient.TargetNamespace` already has the same value "cert-manager", as the value was duplicated I ended up removing the former.

pkg/controller/deployment/credentials_request.go
swghosh: Yes added a new hook: `withSABoundToken`.

pkg/controller/deployment/credentials_request.go
swghosh: Good catch, de-duped.

pkg/controller/deployment/credentials_request.go
swghosh: Refactored to use switch case instead.

pkg/controller/deployment/credentials_request.go
swghosh: Included in latest commit, thanks!

pkg/controller/deployment/deployment_overrides_validation.go
swghosh: Done

swghosh:
/cc @davemulford 
for px-approved label
/cc @xenolinux 
for docs-approved labe;
/cc @xingxingxia 
for qe-approved, could you please approve this one, feel free to test this one out in our next staging build (post-merge)

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,105 @@
...
+				},
+			)
+			deployment.Spec.Template.Spec.Containers[0].VolumeMounts = append(
+				deployment.Spec.Template.Spec.Containers[0].VolumeMounts,
+				corev1.VolumeMount{
+					Name:      cloudCredentialsVolumeName,
+					MountPath: gcpCredentialsDir,
+				},
+			)
+		}
```
thejasn, line 102: Can you refactor this to so that we don't have multiple `append()`. Create a ref for volume and volumeMount outside the switch. 

Also return error on `default` in the switch, we need to catch use cases of other platforms that are using credentials (STS). WDYT?

swghosh: Cool, used:
```
var volume *corev1.Volume
var volumeMount *corev1.VolumeMount
```
and a single set of appends towards the end of the func.

Regarding, `default` case updated it return an error, this would also help discourage users not on AWS/GCP from using that env var for mounting secrets, thanks!

pkg/controller/deployment/credentials_request.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// cloudCredentialsVolumeName is the volume name for mounting
+	// service account (gcp) or credentials (aws) file
+	cloudCredentialsVolumeName = "cloud-credentials"
+)
+
+func withCloudCredentials(secretsInformer coreinformersv1.SecretInformer, infraInformer configinformersv1.InfrastructureInformer, deploymentName, secretName string) func(operatorSpec *v1.OperatorSpec, deployment *appsv1.Deployment) error {
+	// cloud credentials is only required for the controller deployment,
+	// other deployments should be left untouched
+	if deploymentName != "cert-manager" {
```
thejasn, line 35:
```suggestion
	if deploymentName != certmanagerControllerDeployment {
```

swghosh: Updated in latest.

xingxingxia:
> for qe-approved, could you please approve this one, feel free to test this one out in our next staging build (post-merge)

Got. Though, no matter of pre-merge / post-merge test, I'd like to give /qe-approved after dev peers' /lgtm present.

swghosh:
/jira cc-qa

> Got. Though, no matter of pre-merge / post-merge test, I'd like to give qe-approved after dev peers' lgtm present.

@xingxingxia guess now you can add :)

xingxingxia:
Will do post-merge test.
/label qe-approved

PR #124, opened 2023-05-17
Change order of logLevel hook

swghosh:
Updated ordering of `withLogLevel` deployment hook.

> to ensure container args override can take precedence over it,
> helps support users to use different values of --v for each cert-manager component
> at a granular level on top of what level is in spec.logLevel


swghosh:
/label docs-approved
adding this label by myself as @xenolinux is OOO
The necessary doc for this change is being addressed under https://github.com/openshift/openshift-docs/pull/59777/files#r1190702243 which should suffice

swghosh:
/jira cc-qa
@xingxingxia for qe-approved label

xingxingxia:
Will test post merge to validate components log args > logLevel .
/label qe-approved

PR #128, opened 2023-05-26
OCPBUGS-14129: Nil panic fix in validateLabels

swghosh:
Fixes a nil panic caused via deployment_overrides_validation.go due to incorrectly passing value of controllerConfig instead of webhookConfig/caInjectorConfig


swghosh:
cc: @xenolinux for docs-approved label

thejasn:
cc: @davemulford

swghosh:
/cc @davemulford
for px-approved label

PR #135, opened 2023-07-11
Use ubi:9.2 base image for Dockerfile(s)

swghosh:
Bump to ubi:9.2 from 9.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cert-manager-operator
 COPY . .
 RUN make build --warn-undefined-variables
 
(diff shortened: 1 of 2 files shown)


swghosh:
/label docs-approved
this PR doesn't introduce any user-facing changes
/label px-approved
no changes to user experience, our CPaaS pipelines already use UBI base for the operator builds
/label qe-approved
this is essentially a no-op, because in our CI we already started using UBI images a while ago: https://github.com/openshift/release/blob/324ef1c94c8488c835237f68ac48634919e6f555/ci-operator/config/openshift/cert-manager-operator/openshift-cert-manager-operator-master.yaml#L31

PR #136, opened 2023-07-11
OCPBUGS-13830: Add support for overriding container resources

arkadeepsen:
This PR adds the OverrideResources field to DeploymentConfig. This would add the support of overriding the resources fields of the cert-manager, the cert-manager-cainjector, and the cert-manager-webhook deploymemts. Additionally, the resource requests for cert-manager-operator deployment is added.


TrilokGeer:
Can PR description and subject be updated? Maybe https://hypershift-docs.netlify.app/contribute/ is of help.

pkg/controller/deployment/deployment_helper.go
```diff
@@ -76,6 +76,74 @@ func parseArgMap(argMap map[string]string, args []string) {
...
+// provided as override resources.
+func mergeContainerResources(sourceResources corev1.ResourceRequirements, overrideResources corev1.ResourceRequirements) corev1.ResourceRequirements {
+	mergedResourceLimits := mergeContainerResourceList(sourceResources.Limits, overrideResources.Limits)
+	mergedResourceRequests := mergeContainerResourceList(sourceResources.Requests, overrideResources.Requests)
+	mergedResourceClaims := mergeResourceClaims(sourceResources.Claims, overrideResources.Claims)
+
+	return corev1.ResourceRequirements{
+		Limits:   mergedResourceLimits,
+		Requests: mergedResourceRequests,
+		Claims:   mergedResourceClaims,
```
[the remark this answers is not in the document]
arkadeepsen: Makes sense. I had thought as the field is included, better to support the merge functionality. Will remove it.

pkg/controller/deployment/deployment_helper.go
```diff
@@ -76,6 +76,74 @@ func parseArgMap(argMap map[string]string, args []string) {
 	}
 }
 
+// mergeContainerResources merges source container resources with that
+// provided as override resources.
+func mergeContainerResources(sourceResources corev1.ResourceRequirements, overrideResources corev1.ResourceRequirements) corev1.ResourceRequirements {
```
thejasn, line 81: Is it possible to make use of https://github.com/kubernetes/kubernetes/blob/7f6f68fdabc4df88cfea2dcf9a19b2b830f1e647/pkg/api/v1/resource/helpers.go#L377?

swghosh: +1

swghosh: The func signature in this `MergeContainerResourceLimits` fn uses `v1.Container` type and two things to note:
- it doesn't set `hugepage` related resources
- it doesn't set the overriden limit in case the limit was already set in the container (i.e. if we use this and in the future we end up adding resource *limits* to our operator deployment (which is currently not present) we have to change this implementation)

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -664,7 +664,10 @@ spec:
                 image: openshift.io/cert-manager-operator:latest
                 imagePullPolicy: Always
                 name: cert-manager-operator
-                resources: {}
+                resources:
```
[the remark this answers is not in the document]
swghosh: Is the operator container resource requests/limits customizable in any way?
We're setting it to these values here in the CSV which is a default. Asking this because, the original bug report is also about the operator resources and not just for operand.

pkg/controller/deployment/deployment_helper.go
```diff
@@ -76,6 +76,74 @@ func parseArgMap(argMap map[string]string, args []string) {
...
+func mergeContainerResourceList(sourceResourceList corev1.ResourceList, overrideResourceList corev1.ResourceList) corev1.ResourceList {
+	if sourceResourceList == nil && overrideResourceList == nil {
+		return nil
+	}
+
+	destResourceList := corev1.ResourceList{}
+	parseResourceList(destResourceList, sourceResourceList)
+	parseResourceList(destResourceList, overrideResourceList)
+
+	return destResourceList
```
thejasn, line 104: This feels redundant as `ResourceList` is already a mad[resourceName]Quantity. Why can't we just explicitly set values for the `Memory` and `CPU`?

pkg/controller/deployment/deployment_helper.go
thejasn: signature change shouldn't matter to us, but anyways we can't reuse. 

> it doesn't set hugepage related resources
We do not want to support hugepages either since this will involve overriding volumes and volumemounts. So until someone files a bug/rfe. 

> it doesn't set the overriden limit in case the limit was already set in the container (i.e. if we use this and in the future we end up adding resource limits to our operand deployment (which is currently not present) we have to change this implementation)
Yeah, this might be an issue.

pkg/controller/deployment/deployment_helper.go
thejasn:
```suggestion
	if sourceResourceList == nil && overrideResourceList == nil {
		return nil
	}

	destResourceList := corev1.ResourceList{}
	if !sourceResourceList[corev1.ResourceCPU].IsZero() {
	        destResourceList[corev1.ResourceCPU] = sourceResourceList[corev1.ResourceCPU].DeepCopy()
	}
	
	if !sourceResourceList[corev1.ResourceMemory].IsZero() {
	        destResourceList[corev1.ResourceMemory] = sourceResourceList[corev1.ResourceMemory].DeepCopy()
	}

	return destResourceList
```

pkg/controller/deployment/deployment_helper.go
```diff
@@ -76,6 +76,74 @@ func parseArgMap(argMap map[string]string, args []string) {
...
+	return corev1.ResourceRequirements{
+		Limits:   mergedResourceLimits,
+		Requests: mergedResourceRequests,
+		Claims:   mergedResourceClaims,
+	}
+}
+
+// mergeContainerResourceList merges source resource list with that
+// provided as override resource list.
+func mergeContainerResourceList(sourceResourceList corev1.ResourceList, overrideResourceList corev1.ResourceList) corev1.ResourceList {
```
thejasn, line 95: Technically not a merge in case of `ResourceList` since we are basically overriding only here. So might as well rename this.

arkadeepsen: Won't we merge the overriding resources with the source's resources if, say, only limits is provided in the overriding resources and not requests, whereas source's resources has both limits and requests?

If we override then whatever is in the overriding resources, even if either of empty limits or requests, will override the source's limits and requests even if something is already set there.

thejasn: Overriding resources takes priority so we set whatever is set there as the final value. What you are suggesting would be the case if one could update source's resources via a different channel as compared `overrideResources`.

api/operator/v1alpha1/certmanager_types.go
```diff
@@ -77,6 +77,10 @@ type DeploymentConfig struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	OverrideLabels map[string]string `json:"overrideLabels,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +optional
+	OverrideResources corev1.ResourceRequirements `json:"overrideResources,omitempty"`
```
swghosh, line 83: Can we change this type to use a self-created struct that just contains Resources and Limits (essentially stripping out Claims as well, which would be counter-intuitive for discouraging users about adding any Claims as well).

It's better to not rely upon upstream type, rather have a copied one with only the necessary fields, xref: https://github.com/openshift/api/pull/1387#discussion_r1182817836

pkg/controller/deployment/deployment_helper.go
thejasn: We will never have default resources, so yes. And if `overrideResources` is `nil` nothing to do but if `overrideResources` is `{}` then yes that will take precedence.

arkadeepsen: I had a discussion with @TrilokGeer that we might want to do some study and set some default values in the future. For reference IBM's cert-manager operator sets some default values: https://github.com/IBM/ibm-cert-manager-operator/blob/master/controllers/resources/certmanager.go

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
thejasn: We are only setting the `requests` so should be ok, but yes limits need to be customizable. Can be done using `subscription.spec.config` so not an issue.

pkg/controller/deployment/deployment_helper.go
thejasn: For requests we could have something in the future but for limits I'd still be hesitant to push during upgrades especially when majority of the users are using `Automatic` upgrades. This can be viewed as a potential breaking change during upgrades since we will not definitively know at what scale users are using cert-manager.  Limits are usually [frowned upon ](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits)on control plane components and although cert-manager doesn't affect control plane now it could in the future.

thejasn: Now that there is a possibility of adding default `requests` at least, I think we should be okay doing a `nil`/`empty` check before overriding so that we don't override with a nil/empty value for mem/cpu.

swghosh: > Limits are usually [frowned upon ](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits)on control plane components and although cert-manager doesn't affect control plane now it could in the future.

In the future, it is possible that API server endpoints use a cert-manager issued certificate and in that case if the operator/cert-manager operand goes down it could be a SLO problem.  I'd suggest preferring to refrain from adding any resource limits as of now (unless user themselves configures as such: which with this implementation they can do through overrideResources for operand and through Subscription object for the operator), also partly because upstream cert-manager also doesn't enforce these limits yet so we don't know what are good enough sane values for clusters with large certificate usage.

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
swghosh: Ok, we need to document it later (like we did for logLevels).
+doc-ack, cc: @xenolinux

pkg/controller/deployment/deployment_helper.go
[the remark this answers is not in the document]
swghosh: Added a follow-up comment in https://github.com/openshift/cert-manager-operator/pull/136#discussion_r1280991062.

pkg/controller/deployment/deployment_helper.go
arkadeepsen: So, if we set default requests and not limits, however, overriding resources has limits and no requests, do we remove the default requests and use whatever is in the overriding resources? Or do we keep the default resources and use the overriding limits?

arkadeepsen: > In the future, it is possible that API server endpoints use a cert-manager issued certificate and in that case if the operator/cert-manager operand goes down it could be a SLO problem. I'd suggest preferring to refrain from adding any resource limits as of now (unless user themselves configures as such: which with this implementation they can do through overrideResources for operand and through Subscription object for the operator), also partly because upstream cert-manager also doesn't enforce these limits yet so we don't know what are good enough sane values for clusters with large certificate usage.

We might not set it right now, but the implementation should take into account the changes we might have in the future. So if we decide to add some default requests, an empty `overrideResources` will always end up removing that if we decide that it will override and not merge with default resources.

swghosh: Agreed, implementation should be in a way that it takes care of such scenarios.

I would say,
```yaml
originalDeployment:
  resources:
      requests: X
      limits: nil

overrides:
   resources:
     resources: nil
     limits: Y
```

should translate to:
```yaml
finalDeployment:
   resources:
     requests: X
     limits: Y
```
Hope it makes sense.

swghosh: Whereas,
```yaml
originalDeployment:
  resources:
      requests: X
      limits: nil

overrides:
   resources:
     resources: Z
     limits: Y
```

should still translate to:
```yaml
finalDeployment:
   resources:
     requests: Z
     limits: Y
```

swghosh: I was meaning to say, we stick to:
```yaml
finalDeployment:
  resources:
      requests:
        cpu: X
        memory: Y
      limits: nil
```
in such scenarios (which involves merging individual resource types, `iff not nil`).

pkg/controller/deployment/deploment_helper_test.go
```diff
@@ -0,0 +1,398 @@
...
+							Limits: corev1.ResourceList{
+								corev1.ResourceCPU: k8sresource.MustParse("10m"),
+							},
+							Requests: corev1.ResourceList{
+								corev1.ResourceMemory: k8sresource.MustParse("128Mi"),
+							},
+						},
+					},
+				},
+			},
```
swghosh, line 298: Can a add another or a couple test case(s) to test the behaviour we discussed in case of:  https://github.com/openshift/cert-manager-operator/pull/136#discussion_r1281566526

arkadeepsen: Those are added here: https://github.com/openshift/cert-manager-operator/pull/136/files#diff-fe3b58419cf80f35e0dc128fc74a9aa700807ea529635eafa1e44c8d1f39ebd6R23

pkg/controller/deployment/deploment_helper_test.go
```diff
@@ -0,0 +1,398 @@
...
+				},
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    k8sresource.MustParse("40m"),
+					corev1.ResourceMemory: k8sresource.MustParse("64Mi"),
+				},
+			},
+		},
+		{
+			name: "override resources limits and requests merges with source resource limits and requests respectively",
+			sourceResources: corev1.ResourceRequirements{
```
swghosh, line 178: Can you add test cases for testing limits and request as nil explicitly?

Similar to:
```
{
			name: "override resources limits and requests merges with source resource limits and requests respectively",
			sourceResources: corev1.ResourceRequirements{
				Limits: nil,
				Requests: corev1.ResourceList{
					corev1.ResourceCPU: k8sresource.MustParse("10m"),
				},
			},
			overrideResources: v1alpha1.CertManagerResourceRequirements{
				Limits: corev1.ResourceList{
					corev1.ResourceCPU: k8sresource.MustParse("400m"),
				},
				Requests: corev1.ResourceList{
					corev1.ResourceMemory: k8sresource.MustParse("64Mi"),
				},
			},
			expected: corev1.ResourceRequirements{
				Limits: corev1.ResourceList{
					corev1.ResourceCPU: k8sresource.MustParse("400m"),
				},
				Requests: corev1.ResourceList{
					corev1.ResourceCPU:    k8sresource.MustParse("10m"),
					corev1.ResourceMemory: k8sresource.MustParse("64Mi"),
				},
			},
		}
```

pkg/controller/deployment/deployment_overrides_test.go
```diff
@@ -396,3 +400,145 @@ func TestParseArgMap(t *testing.T) {
...
+							Containers: []corev1.Container{
+								{
+									Name: "test-container",
+								},
+							},
+						},
+					},
+				},
+			},
+			overrideFn: func(_ certmanagerinformer.CertManagerInformer, _ string) (v1alpha1.CertManagerResourceRequirements, error) {
```
thejasn, line 430: @arkadeepsen I don't the value running `withContainerResourcesOverrideHook()` in the test if are also passing the override test function. We can ditch this in favour of covering this as part of e2e.

pkg/controller/deployment/deployment_overrides_validation_test.go
```diff
@@ -0,0 +1,739 @@
...
+			}
+
+			// Wait for the informer to get the event.
+			select {
+			case <-certManagerChan:
+			case <-time.After(wait.ForeverTestTimeout):
+				t.Fatal("Informer did not get the added cert manager object")
+			}
+
+			err = withContainerResourcesValidateHook(certManagerInformers, tc.deploymentName)(nil, nil)
```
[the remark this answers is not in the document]
thejasn: We can ditch this in favour of covering this as part of e2e.

thejasn (changes requested):

pkg/controller/deployment/deployment_overrides_validation_test.go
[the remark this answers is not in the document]
arkadeepsen: > @arkadeepsen I believe we are not really testing anything with this function correct me if I'm mistaken. Since you passing nil as args to the returned function which doesn't use any of the arguments in the returned function. I don't see the value here.

The function which is returned doesn't use any of the arguments. Thus passing anything in the call wouldn't make any difference. Also the `withContainerResourcesValidateHook` function is similar to the existing `ValidateHook` functions. 

> We can ditch this in favour of covering this as part of e2e.

The unit test is added to test whether this function is working as expected. We are anyway testing the overall functionality through the e2e test.

pkg/controller/deployment/deployment_overrides_test.go
arkadeepsen: This unit test is for checking whether the `withContainerResourcesOverrideHook` is working correctly or not with the inputs provided. The e2e test will anyway test the overall functionality of the feature.

pkg/controller/deployment/deployment_overrides_validation_test.go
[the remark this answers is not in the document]
thejasn: > The unit test is added to test whether this function is working as expected.

Yes, but if what is the critical business logic that you are testing here? All the functions that are constructed inside `withContainerResourcesValidateHook()` are not dependent on the args and is only dependent on the cluster object. `validateResources` is the actual function you we want to test (and this function is pretty much static if we remove the dep on the informer), I'd suggest to update the scope and make it testable directly. WDYT? This would reduce the boiler plate required for tests and also tests the critical bits.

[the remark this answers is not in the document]
swghosh: please refer to my other comment on https://github.com/openshift/cert-manager-operator/pull/136#discussion_r1285910160, ditto.

pkg/controller/deployment/deploment_helper_test.go
```diff
@@ -0,0 +1,398 @@
...
+					corev1.ResourceMemory: k8sresource.MustParse("64Mi"),
+				},
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    k8sresource.MustParse("40m"),
+					corev1.ResourceMemory: k8sresource.MustParse("64Mi"),
+				},
+			},
+		},
+		{
+			name: "override resources limits and requests merges with source resource limits and requests respectively",
```
swghosh, line 177: small nit: would you mind chaning to:
"override resources replace source limits and requests" instead?

swghosh:
```suggestion
			name: "override resources replace source limits and requests",
```

pkg/controller/deployment/deployment_overrides_validation_test.go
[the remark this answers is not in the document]
arkadeepsen: Ok. I'll change the scope of `validateResources` and add unit test for it. I'll remove the unit test for `withContainerResourcesValidateHook`.

pkg/controller/deployment/deployment_overrides_test.go
arkadeepsen: Ok. I'll remove the unit test for `withContainerResourcesOverrideHook`.

pkg/controller/deployment/deploment_helper_test.go
```diff
@@ -0,0 +1,398 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			actualResources := mergeContainerResources(tc.sourceResources, tc.overrideResources)
+			require.Equal(t, tc.expected, actualResources)
+		})
+	}
+}
+
+func TestGetOverrideResourcesFor(t *testing.T) {
```
arkadeepsen, line 215: @thejasn @swghosh PLMK if this unit test is also redundant. I'll remove this one as well.

pkg/controller/deployment/deploment_helper_test.go
arkadeepsen: This test actually merges. There's not replace of existing values.

pkg/controller/deployment/deploment_helper_test.go
swghosh: I think we shouldn’t remove this one, but rather keep it. Cause, it effectively tests the switch case in the actual func that switches between different deployments (in the past we’ve seen a bug where we accidentally made mistakes in such For functions, so better to retain this).

pkg/controller/deployment/deploment_helper_test.go
swghosh: Cool, then leave it as is.

swghosh:
/cc @davemulford 
for px-approved label
/cc @xingxingxia 
for qe-approved label
(you should be able to test this once we do our first build for 1.12, post-merge test cause this is dev-tested)
/cc @xenolinux 
for docs-approved label

TIA all!

thejasn:
/label px-approved
Adding labels since this is blocking 1.12 release candidate build.

xingxingxia:
/label qe-approved
Glad to see test/e2e/cert_manager_deployment_test.go already covers "cert-manager" deployment. QE decide to do post-merge test for cert-manager-cainjector, and the cert-manager-webhook and operator.

PR #137, opened 2023-07-11
CM-98: Add --enable-certificate-owner-ref as supported cert-manager controller arg

arkadeepsen:
This PR adds the support for the flag `--enable-certificate-owner-ref`. This flag can be enabled by adding it to the `spec.controllerConfig.overrideArgs` field of the CertManager `cluster` object. When the flag is enabled the certificate resource is set as an owner of secret where the tls certificate is stored and the secret will be automatically removed when the certificate resource is deleted.

### DIFF
--- a/test/e2e/cert_manager_deployment_test.go
+++ b/test/e2e/cert_manager_deployment_test.go
@@ -375,7 +375,7 @@ func verifyValidControllerOperatorStatus(t *testing.T, client *certmanoperatorcl
 
 func addValidControlleDeploymentConfig(operator *v1alpha1.CertManager) {
 	operator.Spec.ControllerConfig = &v1alpha1.DeploymentConfig{
-		OverrideArgs: []string{"--dns01-recursive-nameservers=10.10.10.10:53", "--dns01-recursive-nameservers-only"},
+		OverrideArgs: []string{"--dns01-recursive-nameservers=10.10.10.10:53", "--dns01-recursive-nameservers-only", "--enable-certificate-owner-ref"},
 		OverrideEnv: []corev1.EnvVar{
 			{
 				Name:  "HTTP_PROXY",
--- a/pkg/controller/deployment/deployment_overrides_validation.go
+++ b/pkg/controller/deployment/deployment_overrides_validation.go
@@ -43,6 +43,10 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
 		// Application Default Credentials (ADC) per
 		// https://cloud.google.com/docs/authentication/application-default-credentials#search_order
 		"--issuer-ambient-credentials",
+		// Whether to set the certificate resource as an owner of secret where the tls certificate
+		// is stored. When this flag is enabled, the secret will be automatically removed when the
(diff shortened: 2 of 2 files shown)


test/e2e/cert_manager_deployment_test.go
```diff
@@ -375,7 +375,7 @@ func verifyValidControllerOperatorStatus(t *testing.T, client *certmanoperatorcl
 
 func addValidControlleDeploymentConfig(operator *v1alpha1.CertManager) {
 	operator.Spec.ControllerConfig = &v1alpha1.DeploymentConfig{
-		OverrideArgs: []string{"--dns01-recursive-nameservers=10.10.10.10:53", "--dns01-recursive-nameservers-only"},
+		OverrideArgs: []string{"--dns01-recursive-nameservers=10.10.10.10:53", "--dns01-recursive-nameservers-only", "--enable-certificate-owner-ref"},
```
swghosh, line 378: I think we should refrain from adding this flag here, as it would mean that every e2e test would end up using certificate-owner-ref which isn't something we want to test (IMO better not to change default behaviour for now) rather we add e2e(s) specific to this scenario as a follow-up PR to specifically test the effects of applying this flag (similar to discussion we had today).

`"--dns01-recursive-nameservers=10.10.10.10:53", "--dns01-recursive-nameservers-only"` can still remain IMHO because some certificate tests on AWS explicitly needs these flags.

swghosh (approved):
Mostly lgtm, except one small suggestion.

TrilokGeer:
Can PR description and subject be updated? Maybe https://hypershift-docs.netlify.app/contribute/ is of help.

swghosh:
> Can PR description and subject be updated? Maybe https://hypershift-docs.netlify.app/contribute/ is of help.

@arkadeepsen Could you please add a note regarding release note for this PR? So, when we do the 1.12 release, we can share the PR description to docs directly.

cc: @xenolinux

test/e2e/cert_manager_deployment_test.go
arkadeepsen: The function `addValidControlleDeploymentConfig` is only used once inside `TestContainerOverrides` e2e test and not all e2e tests. Also these flags are added once and the validity of the status of the cert-manager operator is checked. After that the flags are removed. Thus this flags do not effect any other tests. 

When we modularize the e2e test, we anyway need to add the flag `--enable-certificate-owner-ref` to check whether it is correctly working or not. Adding it here, IMO, verifies that it is working correctly and makes the PR complete.

Let me know your thoughts.

swghosh: Cool, then we're good to keep it here for now.

swghosh:
/cc @thejasn 
for approved label
/cc @davemulford 
for px-approved label
/cc @xenolinux 
for docs-approved label
/cc @xingxingxia 
for qe-approved label

thejasn:
/approve

@xenolinux We'll need to add a warning/note when using this field, because once this arg is added if we uninstall the operand/operator or delete the certificate this will break network connectivity since the attached secret is deleted.

PR #140, opened 2023-08-09
CM-158: E2E test cases

swghosh:
E2E tests related


test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+		})
+	})
+
+	Context("dns-01 challenge using ambient credentials", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
```
arkadeepsen, line 187: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+		})
+	})
+
+	Context("http-01 challenge using ingress", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-explicit-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
```
arkadeepsen, line 285: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			By("creating a cluster issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer.yaml"), ns.Name)
```
arkadeepsen, line 289: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			By("creating an openshift-hello deployment")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "deployment.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "deployment.yaml"), ns.Name)
```
arkadeepsen, line 293: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			By("creating a service for the deployment openshift-hello")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "service.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "service.yaml"), ns.Name)
```
arkadeepsen, line 297: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+					},
+					TLS: []networkingv1.IngressTLS{{
+						Hosts:      []string{ingress_host},
+						SecretName: "ingress-prod-secret",
+					}},
+				},
+			}
+			ingress, err = loader.KubeClient.NetworkingV1().Ingresses(ingress.ObjectMeta.Namespace).Create(ctx, ingress, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.KubeClient.NetworkingV1().Ingresses(ingress.ObjectMeta.Namespace).Delete(ctx, ingress.ObjectMeta.Name, metav1.DeleteOptions{})
```
arkadeepsen, line 340: Should the error be checked for the this function?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+						Kind: "Issuer",
+					},
+				},
+			}
+
+			_, err = certmanagerClient.CertmanagerV1().Certificates(ns.Name).Create(ctx, cert, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("checking for certificate secret contents")
+			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
```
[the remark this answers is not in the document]
swghosh: This is intentional, checking there is no error returned by this function as PollImmediate gets pre-empted at the return of any error. In this case, the secrets get is the only call which is error-prone, however we expect for the first few minutes for this secret to not be present thus keep retyring with PollImmediate, but in the event of the returning the "secret not found" error from the Get call or the alike, the poll retries would stop which isn't desired.

Instead to avoid confusion, I can use `_ = wait.PollImmediate(...)` WDYT?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+	t := &testing.T{}
+
+	var ctx context.Context
+	var loader library.DynamicResourceLoader
+
+	var routeClient *library.RoutesClient
+	var infraClient *infrav1.ConfigV1Client
+	var certmanagerClient *certmanagerclientset.Clientset
+
+	var appsDomain string
```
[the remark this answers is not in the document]
swghosh: Hmm, moved most of these initializations to suite_test under BeforeAll. However, I still choose to keep the `context.Context` in here.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
arkadeepsen: `wait.PollImmediate` will return an error when the timeout occurs or the function of type `wait.ConditionFunc`returns false or an error. As you are not returning the error of "secret not found" from the function, it will return true or false based on the validity of the secret tls. Thus, the expected output is that the secret should be available within the timeout and it should be valid. Checking the error will validate this, or else we wouldn't know whether finally the correct secret was created or not.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+		Expect(err).NotTo(HaveOccurred(), "Operator is expected to be available")
+	})
+
+	Context("dns-01 challenge using explicit credentials", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-explicit-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
```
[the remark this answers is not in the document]
swghosh: Checking error(s) for the defer call, makes the code look like the following:
```
// could be placed in utils.go
func AckCondition(condFunc func() (bool, error), errHandleFunc func(error)) {
	cond, err := condFunc()
	if !cond {
		errHandleFunc(err)
	}
}
```
with the actual defer call as:
```
defer library.AckCondition(func() (bool, error) {
	return loader.DeleteTestingNS(ns.Name)
}, func(err error) {
})
```

Also, there is slight pain in directly checking the error with the Delete* func(s) from library/utils, because their implementation returns the error of a Get call performed after actual delete of the resource. Irrespective of it's a k8s not found, the error is returned. Currently, that feels bit counter intuitive as it should only return an error when deletion have failed but am not willing to change the whole implementation since it has been there for quite sometime and is well tested in practice.

Let me know if you have an alternate error checking strategy with this or if we could skip the error test in the all the defer delete calls altogether (like the earlier tests did).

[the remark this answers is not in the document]
swghosh: cc: @thejasn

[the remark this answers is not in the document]
thejasn: If we are using defers to **only** delete NS, we skip the error check, seems a bit extra IMO.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			By("checking for certificate secret contents")
+			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
+				secret, _ := loader.KubeClient.CoreV1().Secrets(ns.Name).Get(ctx, certName, metav1.GetOptions{})
+				_, isValid := library.GetTLSConfig(secret)
+				if !isValid {
+					return false, nil
+				}
+				return true, nil
+			})
```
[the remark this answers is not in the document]
swghosh: Yeah, I created two funcs: `waitForCertificateReadiness` and `verifyCertificate` to avoid duplication of such.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,366 @@
...
```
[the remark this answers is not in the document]
swghosh: Yeah, I created two funcs: waitForCertificateReadiness and verifyCertificate to avoid duplication of such.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
swghosh: Thanks, I get your point (I think so), nevertheless realizing that PollImmediate would return an error even when timeout is exceeded, so always good to check the error out from it.
I will ensure to update all instances of err's returned from wait.PollImmediate to error check, after which you could take a re-look!

thejasn:
/hold 
for e2e tests to pass

test/e2e/utils_test.go
```diff
@@ -167,3 +191,163 @@ func verifyDeploymentArgs(k8sclient *kubernetes.Clientset, deploymentName string
...
+		},
+	}
+	payload, err := json.Marshal(patch)
+	if err != nil {
+		return err
+	}
+
+	subscriptionClient := loader.DynamicClient.Resource(subscriptionSchema).Namespace("cert-manager-operator")
+	_, err = subscriptionClient.Patch(ctx, subName, types.MergePatchType, payload, v1.PatchOptions{})
+	return err
```
arkadeepsen, line 352: Better to add this code within retryOnConflict.

test/e2e/utils_test.go
```diff
@@ -167,3 +191,163 @@ func verifyDeploymentArgs(k8sclient *kubernetes.Clientset, deploymentName string
...
+		// return an error, else certificate was renewed properly
+		if !cert.NotAfter.After(*initExpiryTime) {
+			return false, fmt.Errorf("previous expiry time of the certificate cannot be ahead of the current expiry time")
+		}
+
+		wasRenewedAtleastOnce = true
+		return true, nil
+	})
+	return wasRenewedAtleastOnce, err
+}
```
arkadeepsen, line 321: Same here as that of `verifyCertificate`.

arkadeepsen (commented):
Mostly fine, only a few comments from my side.

test/e2e/utils_test.go
swghosh: I ended up removing the .Update() call to certmanager.operator object, because it was incorrectly used here (without a prior .Get, erroring out the update, failing the e2e ultimately).

Instead, it felt more clean to use a .Patch() call with the certmanager object (so we don't have to Get, set the specific fields to nil, and then perform... i.e. more code to maintain for future API changes). @arkadeepsen do you still think I should use `retry.retryOnConflict` with .Patch() calls also (IMHO, I saw in most places it's just paired with Update generally so unsure if Patch also needs it).

@thejasn it would be seemingly like (this is in favour of least maintainability of `resetCertManager` state):
```golang
// delete the operator.openshift.io/v1alpha1 CertManager object,
// it should get created again with default values,
// remove the finalizers before deleting the object
finalizerPatch := []interface{}{
	map[string]interface{}{
		"op":   "remove",
		"path": "/metadata/finalizers",
	},
}
payload, err := json.Marshal(finalizerPatch)
if err != nil {
	return err
}

_, err = client.OperatorV1alpha1().CertManagers().Patch(ctx, "cluster", types.JSONPatchType, payload, metav1.PatchOptions{})
if err != nil {
	return err
}
``` 
T'is the one that works,
- .Update() needs me to set each API field within spec to null/{} manually
- simple .Delete() get stalls after a few repeated calls due to unclean finalizer cleanup sometimes
-  this one works well across multiple repeated runs (I shall wait for e2e to pass and verify the same)

Let me know :)

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,478 @@
...
+	BeforeAll(func() {
+		By("creating Kube clients")
+		ctx = context.Background()
+		loader = library.NewDynamicResourceLoader(ctx, t)
+
+		config, err := library.GetConfigForTest(t)
+		Expect(err).NotTo(HaveOccurred())
+
+		configClient, err = configv1.NewForConfig(config)
+		Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
swghosh: Thanks for pointing out, I did find an inconsistency here. The `Context("ACME Certificate")` has an init for `loader` while `Context("Self-signed Certificate")` doesn't, but loader is used by both. This can quickly get confusing and invite nil pointer refs in the future. 

I'll rectify and move these initialization in the `BeforeSuite` irrespective of their usage in specific Context(s) which should help naturally keep the init and the var declaration together atleast.

test/e2e/utils_test.go
arkadeepsen: With `Patch` `retry.retryOnConflict` is probably not required.

swghosh: The patch + delete isn't working well, the operator status on the certmanager object doesn't reconcile back properly, ends up failing the test.
So I'll switch back to Get + Update with a retry.retryOnConflict wrapped around.

swghosh:
/cc @xingxingxia 
for qe-approved label

self adding docs-approved, px-approved since changes in these PR are e2e tests functionality for which have already been added in former releases, doc is also covered.
/label docs-approved
/label px-approved

test/e2e/utils_test.go
swghosh: Done.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
swghosh: Updated!

test/e2e/certificates_test.go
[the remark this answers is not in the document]
swghosh: All updated now :)

test/e2e/utils_test.go
```diff
@@ -66,39 +79,58 @@ func verifyOperatorStatusCondition(client *certmanoperatorclient.Clientset, cont
...
+		updatedOperator.Spec.OperatorSpec = opv1.OperatorSpec{
+			ManagementState: opv1.Managed,
 		}
 
-		if !hasOverride {
-			return nil
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			_, err = client.OperatorV1alpha1().CertManagers().Update(context.TODO(), updatedOperator, v1.UpdateOptions{})
+			return err
+		})
```
arkadeepsen, line 105: I think the get call needs to be inside `retry.RetryOnConflict `, or else the conflict may not get resolved.

swghosh: Cool, lemme update it then.

swghosh: I'm also verifying locally if I need to check the `*Config`'s first before setting them to `nil` i.e. if kube-apiserver returns an error if update has 0 changes in it. Will modify accordingly if required as well.

swghosh: > if update has 0 changes in it

just to confirm, this isn't a problem, apiserver can handle such updates without returning errors.

swghosh: > I think the get call needs to be inside retry.RetryOnConflict , or else the conflict may not get resolved.

@arkadeepsen Updated it now!

test/e2e/utils_test.go
```diff
@@ -66,39 +79,63 @@ func verifyOperatorStatusCondition(client *certmanoperatorclient.Clientset, cont
...
+			}
+			return false, err
 		}
 
-		updatedOperator := operator.DeepCopy()
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			operator, err := client.OperatorV1alpha1().CertManagers().Get(ctx, "cluster", v1.GetOptions{})
+			if err != nil {
+				return err
+			}
```
arkadeepsen, line 98: It would make more sense if the `wait.PollImmediate` is called inside `retry.RetryOnConflict` to get the most recent cert-manager object. Only the get call can be added inside `wait.PollImmediate`, and once the object is fetched the other actions can be performed on it, including the update.

swghosh: Understood, let me update.

swghosh: Done.

test/e2e/overrides_test.go
```diff
@@ -140,13 +140,15 @@ var _ = Describe("Overrides test", Ordered, func() {
 		})
 	})
 
-	AfterAll(func() {
+	AfterEach(func() {
 		By("Reset cert-manager state")
 		err := resetCertManagerState(context.Background(), certmanageroperatorclient, loader)
 		Expect(err).NotTo(HaveOccurred())
+	})
```
arkadeepsen, line 147: This change seems to be an overkill. The`BeforeEach` section already has the call to the `resetCertManagerState`. Adding the same in `AfterEach` section will essentially increase the calls to the api-server and the reconciles.

swghosh: Yeah I understand, but without the `resetCertManagerState` in both BeforeEach and AfterEach, what ends up happening is that thre is unclean state of the operator by the end of some test, which affects the other tests and fails out.
eg. https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1694727559633178624, https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-alpha-e2e-operator/1694727559549292544 it's reproducible across runs; adding this resolves it.

swghosh: That doesn't seem to have resolved the issue I think.
Webhook still goes down and test flakes altogether.
Recent run reports webhook failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1694970785124323328.

arkadeepsen: > Yeah I understand, but without the `resetCertManagerState` in both BeforeEach and AfterEach, what ends up happening is that thre is unclean state of the operator by the end of some test, which affects the other tests and fails out. eg. https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1694727559633178624, https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-alpha-e2e-operator/1694727559549292544 (reconcile failures in: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/140/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1694727559633178624/artifacts/e2e-operator/gather-extra/artifacts/pods/cert-manager-operator_cert-manager-operator-controller-manager-f6b7fd94d-7vdsw_cert-manager-operator.log) it's reproducible across runs; adding this resolves it, and tests pass.

The Override tests shouldn't effect the Certificates test as both the ci runs the Certificates test is run before the Overrides tests. So this seems unrelated to the additional call to the `resetCertManagerState`.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,463 @@
...
+		By("creating Kube clients")
+		ctx = context.Background()
+
+		baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(baseDomain).NotTo(BeEmpty())
+		appsDomain = "apps." + baseDomain
+
+		By("adding override args for dns-01 private zone passthrough")
+		err = addOverrideArgs(certmanageroperatorclient, operatorName, []string{
```
arkadeepsen, line 40: The 2nd argument should be the controller name to which you want the arguments to get added to rather than the operator name. Also you are not waiting for it to trickle down to the deployments. This is probably why you are seeing the webhook endpoint issue.

test/e2e/overrides_test.go
swghosh: Yeah, I checked those logs. certificates_test was executed before overrides_test and all the steps in overrides_test did pass.
However, I found that TestContainerOverrides fn (the old non-ginko test) is always being executed before TestAll (the new ginkgo suite) and there's possibility for that to have caused the regressive behaviour not cleaning up it's own overrides.

I guess this isn't cleaning up properly,
```go

func TestContainerOverrides(t *testing.T) {
         // ...

        // reset operator spec
	operator, err = certmanageroperatorclient.OperatorV1alpha1().CertManagers().Get(ctx, "cluster", metav1.GetOptions{})
	require.NoError(t, err)

	updatedOperator = operator.DeepCopy()
	updatedOperator.Spec = v1alpha1.CertManagerSpec{
		OperatorSpec: opv1.OperatorSpec{
			ManagementState: opv1.Managed,
		},
	}
	_, err = certmanageroperatorclient.OperatorV1alpha1().CertManagers().Update(ctx, updatedOperator, metav1.UpdateOptions{})
	require.NoError(t, err)
}
```

swghosh: Something is still wrong (maybe with TestContainerOverrides func), as with the addOverrides/resetCertManager removed from certificates_test also resulted in the same "webhook unavailable error".

test/e2e/certificates_test.go
arkadeepsen: Mostly LGTM, apart from this incorrect variable name for the cert-manager deployment.

swghosh: I've updated the variable in latest commit (pushing it...)

test/e2e/overrides_test.go
swghosh: Replaced the delete with `resetCertManagerState` func call now. This should resolve it!
Let's wait for e2e to pass one final time.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,463 @@
...
+		By("creating Kube clients")
+		ctx = context.Background()
+
+		baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(baseDomain).NotTo(BeEmpty())
+		appsDomain = "apps." + baseDomain
+
+		By("adding override args for dns-01 private zone passthrough")
+		err = addOverrideArgs(certmanageroperatorclient, certmanagerControllerDeployment, []string{
```
arkadeepsen, line 40:
```suggestion
		err = addOverrideArgs(certmanageroperatorclient, certManagerControllerDeployment, []string{
```
Incorrect variable name

swghosh: Yeah, just now updated the same.

test/e2e/cert_manager_deployment_test.go
```diff
@@ -73,21 +73,23 @@ func TestACMECertsIngress(t *testing.T) {
...
+	defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "service.yaml"), ns.Name)
+
+	configClient, err := configv1.NewForConfig(config)
 	require.NoError(t, err)
+	baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
+	require.NoError(t, err)
+	appsDomain := "apps." + baseDomain
 
-	ingress_host := "hey." + strings.Join(strings.Split(route.Spec.Host, ".")[1:], ".")
+	ingress_host := "eaic." + appsDomain
```
xingxingxia, line 92: To be robust, need it check if the total lengh would be greater than 64 and if yes, make it shorter?

test/e2e/testdata/credentials/credentialsrequest_aws.yaml
```diff
@@ -0,0 +1,28 @@
...
+        resource: arn:aws:route53:::hostedzone/*
+      - action:
+          - route53:ListHostedZonesByName
+        effect: Allow
+        resource: "*"
+  secretRef:
+    name: aws-creds
+    namespace: cert-manager
+  serviceAccountNames:
+    - cert-manager
```
xingxingxia, line 28: Minor issue, it is noeol

test/e2e/cert_manager_deployment_test.go
swghosh: Since, this is an e2e test, I don't think if I need to check the length in my own test code. Because, the cert-manager webhook running on the cluster during the test will always verify the length of the dns name and if it exceeds, it will error out the Create and essentially fail the test, so it's already taken care of in the e2e way, I'd say.

PR #159, opened 2023-10-30
[cert-manager-1.11] CM-221: Bump deps for consuming CVE-2023-44487 fixes

thejasn:
- Bump kube-rbac-proxy to v0.15.0 (also disable http2)
- Bump library-go and its deps (k8s)
- Bump kustomize to v5

PR #161, opened 2023-10-30
[cert-manager-1.11] Bump golang to 1.20

thejasn:
- Bump builder to golang-1.20-openshift-4.14
- Update go.mod to go=1.20


swghosh:
/cc @xingxingxia @lunarwhite 
for qe-approved label
We need go 1.20 in order to bump library-go in #159 which has the remediated CVE.

PR #166, opened 2023-11-08
CM-220: automate test case OCP-65035 on gcp

geliu2016:
@xingxingxia ^_^

### DIFF
--- a/test/e2e/certificates_test.go
+++ b/test/e2e/certificates_test.go
@@ -18,6 +18,7 @@ import (
 	v1 "github.com/cert-manager/cert-manager/pkg/apis/acme/v1"
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	certmanagermetav1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
+	configv1 "github.com/openshift/api/config/v1"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -26,12 +27,13 @@ import (
 var _ = Describe("ACME Certificate", Ordered, func() {
 	var ctx context.Context
 	var appsDomain string
+	var baseDomain string
 
 	BeforeAll(func() {
 		By("creating Kube clients")
 		ctx = context.Background()
-
-		baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
+		var err error
+		baseDomain, err = library.GetClusterBaseDomain(ctx, configClient)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(baseDomain).NotTo(BeEmpty())
 		appsDomain = "apps." + baseDomain
@@ -172,7 +174,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 	})
 
 	Context("dns-01 challenge using ambient credentials", func() {
-		It("should obtain a valid LetsEncrypt certificate", func() {
+		It("should obtain a valid LetsEncrypt certificate in AWS mint/passthrough cluster", func() {
 
 			By("creating a test namespace")
 			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
@@ -193,7 +195,8 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			By("setting cloud credential secret name in subscription object")
-			err = patchSubscriptionWithCloudCredential(ctx, loader)
+			credentialSecret := "aws-creds"
+			err = patchSubscriptionWithCloudCredential(ctx, loader, credentialSecret)
 			Expect(err).NotTo(HaveOccurred())
 
 			By("getting AWS zone from Infrastructure object")
@@ -264,6 +267,84 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			By("checking for certificate validity from secret contents")
+			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough clusters-65035", func() {
+
+			By("Getting Infrastructure object")
+			infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if infra.Status.PlatformStatus.Type != configv1.GCPPlatformType {
+				Skip("Skip this case for current plaform, it's not gcp.")
+			}
+
+			By("Check cloud credential in cluster")
+			_, err = loader.KubeClient.CoreV1().Secrets("kube-system").Get(ctx, "gcp-credentials", metav1.GetOptions{})
+			if err != nil {
+				Skip("Skipping for the cluster without credential in cluster")
+			}
+
+			By("Creating CredentialsRequest object")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "credentials", "credentialsrequest_gcp.yaml"), "")
+
+			By("Waiting for cloud secret to be available")
+			// The name is defined cloud credential by the testdata YAML file.
+			credentialSecret := "gcp-credentials"
+			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
+				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, credentialSecret, metav1.GetOptions{})
+				if err != nil {
+					return false, nil
+				}
+				return true, nil
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Configure cert-manager to use credential, setting this credential secret name in subscription object")
+			err = patchSubscriptionWithCloudCredential(ctx, loader, credentialSecret)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Creating new certificate ClusterIssuer")
+			// The name is defined by the testdata YAML file clusterissuer_gcp.yaml
+			clusterIssuerName := "acme-dns01-clouddns-ambient"
+			gcpProjectId := infra.Status.PlatformStatus.GCP.ProjectID
+			Expect(gcpProjectId).NotTo(Equal(""))
+			replaceStrMap := map[string]string{
+				"PROJECT_ID": gcpProjectId,
+			}
+			loadFileAndReplaceStr := func(fileName string) ([]byte, error) {
+				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
+				return []byte(fileContentsStr), err
+			}
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("Creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01-65035")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("Creating new certificate")
+			randomString := randomStr(3)
+			replaceStrMap = map[string]string{
+				"RANDOM_STR": randomString,
+				"DNS_NAME": baseDomain,
+			}
+			loadFileAndReplaceStr = func(fileName string) ([]byte, error) {
+				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
+				return []byte(fileContentsStr), err
+			}
+			// The name is defined by the testdata YAML file certificate_gcp.yaml
+			certName := "cert-with-acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), ns.Name)
+
+			By("Waiting for certificate to get ready")
+			err = waitForCertificateReadiness(ctx, certName, ns.Name)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("checking for certificate validity from secret contents")
+			certDomain := randomString + "." + baseDomain
 			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
 			Expect(err).NotTo(HaveOccurred())
 		})
--- a/test/e2e/testdata/acme/certificate_gcp.yaml
+++ b/test/e2e/testdata/acme/certificate_gcp.yaml
@@ -1,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: cert-with-acme-dns01-clouddns-ambient
+spec:
+  secretName: cert-with-acme-dns01-clouddns-ambient
+  issuerRef:
+    # The issuer created previously
+    kind: ClusterIssuer
+    name: acme-dns01-clouddns-ambient
+  dnsNames:
+    - RANDOM_STR.DNS_NAME
+    - '*.RANDOM_STR.DNS_NAME'
--- a/test/e2e/testdata/acme/clusterissuer_gcp.yaml
+++ b/test/e2e/testdata/acme/clusterissuer_gcp.yaml
@@ -1,0 +1,15 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: acme-dns01-clouddns-ambient
+spec:
+  acme:
+    preferredChain: ""
+    privateKeySecretRef:
+      name: issuer-acme-letsencrypt
+    server: https://acme-staging-v02.api.letsencrypt.org/directory
+    solvers:
+    - dns01:
+        cloudDNS:
+          # The ID of the GCP project
+          project: PROJECT_ID
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -4,9 +4,11 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"strings"
 	"sync"
 	"time"
@@ -299,7 +301,7 @@ func getCertManagerOperatorSubscription(ctx context.Context, loader library.Dyna
 // patchSubscriptionWithCloudCredential uses the k8s dynamic client to patche the only Subscription object
 // in the cert-manager-operator namespace to inject CLOUD_CREDENTIALS_SECRET_NAME="aws-creds" env
 // into its spec.config.env
-func patchSubscriptionWithCloudCredential(ctx context.Context, loader library.DynamicResourceLoader) error {
+func patchSubscriptionWithCloudCredential(ctx context.Context, loader library.DynamicResourceLoader, credentialSecret string) error {
 	subName, err := getCertManagerOperatorSubscription(ctx, loader)
 	if err != nil {
 		return err
@@ -311,7 +313,7 @@ func patchSubscriptionWithCloudCredential(ctx context.Context, loader library.Dy
 				"env": []interface{}{
 					map[string]interface{}{
 						"name":  "CLOUD_CREDENTIALS_SECRET_NAME",
-						"value": "aws-creds",
+						"value": credentialSecret,
 					},
 				},
 			},
@@ -398,4 +400,28 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
 		// certificate was renewed atleast once
 		return true, nil
 	})
+}
+
+// create randomized string
+func randomStr(size int) string {
+	char := "abcdefghijklmnopqrstuvwxyz0123456789"
+	rand.NewSource(time.Now().UnixNano())
+	var s bytes.Buffer
+	for i := 0; i < size; i++ {
+		s.WriteByte(char[rand.Int63()%int64(len(char))])
+	}
+	return s.String()
+}
+
+// replace string in bytes of ReadFile
+func replaceStrInFile(replaceStrMap map[string]string, fileName string) ([]byte, error) {
+	bytes, err := testassets.ReadFile(fileName)
+	if err != nil {
+		return nil, err
+	}
+	fileContentsStr := string(bytes)
+	for k, v := range replaceStrMap {
+		fileContentsStr = strings.ReplaceAll(fileContentsStr, k, v)
+	}
+	return []byte(fileContentsStr), nil
 }
--- a/test/e2e/testdata/credentials/credentialsrequest_gcp.yaml
+++ b/test/e2e/testdata/credentials/credentialsrequest_gcp.yaml
@@ -1,0 +1,16 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  name: cert-manager
+  namespace: openshift-cloud-credential-operator
+spec:
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/dns.admin
+  secretRef:
+    name: gcp-credentials
+    namespace: cert-manager
+  serviceAccountNames:
+  - cert-manager
--- a/test/library/dynamic_resources.go
+++ b/test/library/dynamic_resources.go
@@ -79,11 +79,10 @@ func (d DynamicResourceLoader) do(do doFunc, assetFunc func(name string) ([]byte
 
 	var dri dynamic.ResourceInterface
 	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
-		require.NotEmpty(d.t, unstructuredObj.GetNamespace(), "Namespace can not be empty!")
-
 		if overrideNamespace != "" {
 			unstructuredObj.SetNamespace(overrideNamespace)
 		}
+		require.NotEmpty(d.t, unstructuredObj.GetNamespace(), "Namespace can not be empty!")
 		dri = d.DynamicClient.Resource(mapping.Resource).Namespace(unstructuredObj.GetNamespace())
 	} else {
 		dri = d.DynamicClient.Resource(mapping.Resource)


xingxingxia:
/retitle CM-220: [WIP] automate test case OCP-65035 on gcp

swghosh:
PS: we need to consider using ginkgo filter test cases for this PR to be merged successfully. As our current e2e tests make use of an AWS CI cluster only and since we're planning to add new test cases for other cloud provider clusters, filtering / ensuring the test suites can be run separately would be required.

test/e2e/certificates_test.go
```diff
@@ -354,6 +356,51 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+	
+	Context("OCP-65035-CLOUD_CREDENTIALS_SECRET_NAME works well in mint/passthrough clusters", func() {
+		FIt("should obtain a valid LetsEncrypt certificate", func() {
+			
+			By("getting Infrastructure object")
+			infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			fmt.Printf("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&infra.Status.PlatformStatus.Type=%v, configv1.AWSPlatformType=%v",infra.Status.PlatformStatus.Type,configv1.AWSPlatformType)
+			if infra.Status.PlatformStatus.Type != configv1.AWSPlatformType {	
+				Skip("Skip this case for current plaform, it's not aws.")
```
xingxingxia, line 368:
```suggestion
			if infra.Status.PlatformStatus.Type != configv1.<GCP>Type {	
				Skip("Skip this case for current plaform, it's not GCP.")
```

test/e2e/certificates_test.go
```diff
@@ -5,9 +5,9 @@ package e2e
 
 import (
 	"context"
+	"fmt"
```
lunarwhite, line 8:
```suggestion
```
In failed job [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/166/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1732239696152498176/artifacts/e2e-operator/test/build-log.txt): `test/e2e/certificates_test.go:8:2: "fmt" imported and not used`

test/e2e/certificates_test.go
```diff
@@ -354,6 +356,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			By("creating new certificate ClusterIssuer")
+			clusterIssuerName := "acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("creating a test namespace")
+                        ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
+                        Expect(err).NotTo(HaveOccurred())
+                        defer loader.DeleteTestingNS(ns.Name)
```
lunarwhite, line 390:
```suggestion
			By("creating a test namespace")
			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
			Expect(err).NotTo(HaveOccurred())
			defer loader.DeleteTestingNS(ns.Name)
```

just nit: need to align space/tab

geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			By("getting Infrastructure object")
+			infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if infra.Status.PlatformStatus.Type != configv1.GCPPlatformType {
+				Skip("Skip this case for current plaform, it's not gcp.")
+			}
+
+			By("Check cloud credential in cluster")
+			_, err = loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "gcp-credentials", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
```
lunarwhite, line 371: we should do this after `By("creating CredentialsRequest object")`, and need apply a `wait-poll` like: https://github.com/openshift/cert-manager-operator/blob/46069515b0bda3a5f51ba88705095a089258d07b/test/e2e/certificates_test.go#L185-L193

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			By("creating new certificate ClusterIssuer")
+			clusterIssuerName := "acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
```
lunarwhite, line 389: namespace created but didn't be used to create certificate in `By("creating new certificate")`.

lunarwhite (commented):
thanks for helping auto! review are still in progress...

test/e2e/certificates_test.go
geliu2016: @lunarwhite Did you confuse the credential and credentialrequest? get gcp-credential is first step in test case [ocp-65035](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitems/testcase?query=id%3AOCP%5C-65035), the 2nd step is create CredentialsRequest

geliu2016:
@xingxingxia BIG appreciate to you if you may review/comment this pr in time.

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
+
+	Context("OCP-65035-CLOUD_CREDENTIALS_SECRET_NAME works well in mint/passthrough clusters", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
```
xingxingxia, line 360: @geliu2016 from the output of either your local test or the PR's bottom's e2e job "ci/prow/e2e-operator", you can see the scenarios' titles are composed by strings in Describe() Context() It():
```
Cert Manager Suite: [It] ACME Certificate dns-01 challenge using ambient credentials should obtain a valid LetsEncrypt certificate
Cert Manager Suite: [It] ACME Certificate OCP-65035-CLOUD_CREDENTIALS_SECRET_NAME works well in mint/passthrough clusters should obtain a valid LetsEncrypt certificate
```
Existing scenario title `ACME Certificate dns-01 challenge using ambient credentials should obtain a valid LetsEncrypt certificate` from the combined Describe() Context() It() strings is a good human-readable title. But your title `ACME Certificate OCP-65035-CLOUD_CREDENTIALS_SECRET_NAME works well in mint/passthrough clusters should obtain a valid LetsEncrypt certificate` from the combined Describe() and your Context() It() strings is a little not much human-readable. May need to update your Context() It() strings to compose a better scenario title.

xingxingxia: You can BTW therefore update existing scenario title `ACME Certificate dns-01 challenge using ambient credentials should obtain a valid LetsEncrypt certificate` to be `ACME Certificate dns-01 challenge using ambient credentials should obtain a valid LetsEncrypt certificate in AWS non-sts cluster` so that above two scenarios are clear for new readers when they just read the titles.

test/e2e/utils_test.go
```diff
@@ -227,7 +227,7 @@ func getCertManagerOperatorSubscription(ctx context.Context, loader library.Dyna
 // patchSubscriptionWithCloudCredential uses the k8s dynamic client to patche the only Subscription object
 // in the cert-manager-operator namespace to inject CLOUD_CREDENTIALS_SECRET_NAME="aws-creds" env
 // into its spec.config.env
-func patchSubscriptionWithCloudCredential(ctx context.Context, loader library.DynamicResourceLoader) error {
+func patchSubscriptionWithCloudCredential(ctx context.Context, loader library.DynamicResourceLoader, credentialValue string) error {
```
xingxingxia, line 230: @geliu2016 Because we are contributing upstream, even if for minor thing like variable name, it is better to be precise. Here, it is not a credential "value", it is just a secret "name", so the variable `credentialValue` is not much precise. How about "credentialSecret"?
Same for other places `credentialValue := "aws-creds"`

test/e2e/certificates_test.go
xingxingxia, line 371: @geliu2016 , like we do for AWS https://github.com/openshift/openshift-tests-private/blob/7b0a42543a777413c09289ea1d95b32fcacb06b5/test/extended/apiserverauth/certmanager.go#L56 , here we need to skip, instead of assert `Expect(err).NotTo(HaveOccurred())`.

xingxingxia: @geliu2016, @lunarwhite is right. Need to update as per @lunarwhite 's `wait` example. Above `wait` example is from Dev's existing code for the AWS case. Just click above link " [cert-manager-operator/test/e2e/certificates_test.go Lines 185 to 193 in 4606951](https://github.com/openshift/cert-manager-operator/blob/46069515b0bda3a5f51ba88705095a089258d07b/test/e2e/certificates_test.go#L185-L193) ".

test/e2e/certificates_test.go
xingxingxia: @geliu2016 , you are right in mapping test case, but @lunarwhite's is also right. Per Dev's code examples for `loader.CreateFromFile`, Dev always gives a namespace variable `ns.Name` as the 3rd argument, for example `loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "...", "...certificate.yaml"), ns.Name)`, but your `loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), "")` does not use your created namespace when you create the certificate.

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			By("creating CredentialsRequest object")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "credentials", "credentialsrequest_gcp.yaml"), "")
+
+			By("Configure cert-manager to use credential, setting this credential secret name in subscription object")
+			credentialValue := "gcp-credentials"
+			err = patchSubscriptionWithCloudCredential(ctx, loader, credentialValue)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating new certificate ClusterIssuer")
+			clusterIssuerName := "acme-dns01-clouddns-ambient"
```
xingxingxia, line 382: The name of the ClusterIssuer is determined by the `clusterissuer_gcp.yaml` file. Therefore the variable `clusterIssuerName` should be retrieved from the file instead of being assigned the value.

Same issue for the variable `certName := "cert-with-acme-dns01-clouddns-ambient"`.

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,50 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("creating new certificate")
+			certName := "cert-with-acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().Certificates(ns.Name).Delete(ctx, certName, metav1.DeleteOptions{})
```
xingxingxia, line 394: @geliu2016 , Certificate is a namespace-scoped resource. As long as the namespace of your certificate is ensured to be deleted, resources within the namespace need not be separately cleaned using `defer`.

test/e2e/testdata/acme/certificate_gcp.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: cert-with-acme-dns01-clouddns-ambient
+spec:
+  secretName: cert-with-acme-dns01-clouddns-ambient
+  issuerRef:
+    # The issuer created previously
+    kind: ClusterIssuer
+    name: acme-dns01-clouddns-ambient
+    dnsNames:
+    - xxia-test.qe1.gcp.devcluster.openshift.com
+    - '*.xxia-test.qe1.gcp.devcluster.openshift.com'
```
xingxingxia, line 13: @geliu2016 To avoid failures again like https://issues.redhat.com/browse/OCPQE-16689 , we need to use random string instead of hard code in auto.

xingxingxia:
@geliu2016 appreciated a lot for the auto! Was on other work, now finished reveiw, please check comments, feel free to discuss whenever needed.

test/e2e/certificates_test.go
geliu2016: @xingxingxia regarding link " [cert-manager-operator/test/e2e/certificates_test.go Lines 185 to 193 in 4606951](https://github.com/openshift/cert-manager-operator/blob/46069515b0bda3a5f51ba88705095a089258d07b/test/e2e/certificates_test.go#L185-L193) ", you may find there is not verification after:
By("creating CredentialsRequest object")
loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "credentials", "credentialsrequest_aws.yaml"), "")

and the wait.PollImmediate is check "Secrets("cert-manager").Get(ctx, "aws-creds", metav1.GetOptions{})", we also have this secret check at beginning:

By("Check cloud credential in cluster")
_, err = loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "gcp-credentials", metav1.GetOptions{})

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/utils_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: Updated.

test/e2e/certificates_test.go
geliu2016: Is there any difference/shortage between retrieved from file and assigning value based on the file which assigned it already? in my view, retrieved from file will add code cost.

test/e2e/certificates_test.go
geliu2016: GOOD CATCH!

xingxingxia:
@geliu2016 could you help check the bottom job showing failure? Thanks.
`ci/prow/e2e-operator` https://github.com/openshift/cert-manager-operator/commit/97f73ae00f373824e90a465b0995a26b73699232          [link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cert-manager-operator/166/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1736654543279624192)

test/e2e/testdata/acme/certificate_gcp.yaml
geliu2016: updated.

geliu2016:
@xingxingxia @lunarwhite could you pls take a quick review? thx

test/e2e/certificates_test.go
```diff
@@ -5,19 +5,19 @@ package e2e
...
 	"context"
-	"path/filepath"
-	"time"
-
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"path/filepath"
+	"time"
```
xingxingxia, line 13: @geliu2016 not sure why the two import lines are moved? I did not care this in last review, but after recently receiving @swghosh 's https://github.com/openshift/cert-manager-operator/pull/168#discussion_r1431238608 , k8s imports, openshift imports and go-lang built-in imports are better to be in separate blocks?

test/e2e/certificates_test.go
xingxingxia: @geliu2016 , no, they are different things.
The beginning checks a `system` secret within `kube-system` namespace. But here we need check a secret defined in your credentialsrequest to be generated within the `cert-manager` namespace. Please check carefully the link again, it shows `("cert-manager")` namespace:
```
_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "aws-creds", metav1.GetOptions{})
```

test/e2e/certificates_test.go
xingxingxia: @geliu2016 sorry if I did not say clear. Per above explanation about "combined", it is expected to update as below:
```
	Context("dns-01 challenge using ambient credential", func() {
		It("should obtain a valid LetsEncrypt certificate in AWS non-sts cluster", func() {
... // Original Dev's automated aws case.
		})

		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough clusters-65035", func() {
... // Your automated gcp case.
	})
```

test/e2e/certificates_test.go
xingxingxia: As talked offline, keep as is, unless Dev shares ideas.

xingxingxia:
Thanks for update. Due to EOD, today reviewed only part. Will continue review next workday.

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,65 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			err = patchSubscriptionWithCloudCredential(ctx, loader, credentialSecret)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating new certificate ClusterIssuer")
+			clusterIssuerName := "acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
```
xingxingxia, line 389: This seems copied from Dev's existing scenario. It should change to use a different name for your scenario. (Did not notice this. Today found it)

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,65 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+				err = replaceStr(randomString, "RANDOM_STR", certFile)
+	                        Expect(err).NotTo(HaveOccurred())
+			}()
+			randomString1 := randomStr(5)
+			err = replaceStr("RANDOM_STR1", randomString1, certFile)
+                        Expect(err).NotTo(HaveOccurred())
+			defer func() {
+                                err = replaceStr(randomString1, "RANDOM_STR1", certFile)
+                                Expect(err).NotTo(HaveOccurred())
+                        }()
```
xingxingxia, line 408: - Are the two `defer`s needed? Seems not needed. I do not get why they are needed.
- Even if the `defer` is needed, `replaceStr(randomString, "RANDOM_STR"` can cause problem when your randomString happens to be some string matching other string in your YAML, e.g. `names`, `amesp` et al can make your YAML file's `namespace` string broken.
- Your YAML file needs not RANDOM_STR1, just one RANDOM_STR is enough:
```
    - RANDOM_STR.qe1.gcp.devcluster.openshift.com
    - '*.RANDOM_STR.qe1.gcp.devcluster.openshift.com'
```

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,65 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+		It("should obtain a valid LetsEncrypt certificate", func() {
+
+			By("getting Infrastructure object")
+			infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if infra.Status.PlatformStatus.Type != configv1.GCPPlatformType {
+				Skip("Skip this case for current plaform, it's not gcp.")
+			}
+
+			By("Check cloud credential in cluster")
```
xingxingxia, line 369: - Sometime it is initial lower case do-ing format `By("getting`, `By("creating`, sometimes it is initial upper case verb format `By("Check`, `By("Configure`. This seems to make inconsistency in RH public repo. It is better to unify format.
- Same for `("Skip ` and `"Skipping `.

test/e2e/certificates_test.go
```diff
@@ -354,6 +355,65 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			_, err = loader.KubeClient.CoreV1().Secrets("kube-system").Get(ctx, "gcp-credentials", metav1.GetOptions{})
+			if err != nil {			
+				Skip("Skipping for the cluster without credential in cluster")
+			}
+
+			By("creating CredentialsRequest object")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "credentials", "credentialsrequest_gcp.yaml"), "")
+
+			By("Configure cert-manager to use credential, setting this credential secret name in subscription object")
+			credentialSecret := "gcp-credentials"
```
xingxingxia, line 379: Readers reading this go code may be confused about where this `"gcp-credentials"` comes from. To make it read clear, it is better to add a comment:
```
// The name is defined by the testdata YAML file.
```
Same for `clusterIssuerName := "acme-dns01-clouddns-ambient"` and `certName := "cert-with-acme-dns01-clouddns-ambient"`.

xingxingxia:
Done review. Thanks!

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: Updated  RANDOM_STR1 to  RANDOM_STR with only one defer needed.
and regarding to random str: RANDOM_STR, I adjusted it size to 7 chars to low the risk of same name conflict, 
I think regarding to 7 char long random str, the risk of same name conflict with other str in file is very very very low.

test/e2e/certificates_test.go
geliu2016: Updated.

test/e2e/certificates_test.go
geliu2016: updated.

geliu2016:
@xingxingxia could you please take a view again? thx

geliu2016:
@xingxingxia could you please take a quick review? thx

test/e2e/utils_test.go
```diff
@@ -327,3 +330,26 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
...
+	var s bytes.Buffer
+	for i := 0; i < size; i++ {
+		s.WriteByte(char[rand.Int63()%int64(len(char))])
+	}
+	return s.String()
+}
+
+// replace string in file
+func replaceStr(oldStr string, newStr string, fileName string) error {
+	f, err := ioutil.ReadFile(fileName)
```
swghosh, line 347: We should read from `testassets.ReadFile` here instead of `ioutil.ReadFile` in order to ensure that this file can be read at runtime directly from `testassets` (embedded FS) instead of having to require the file be present at test/run time.

Alternatively, it should also be good to take the file reader function as an argument to this function something like `func replaceStr(oldStr string, newStr string, assetFunc func(name string) ([]byte, error), fileName string) { ... }` very similar to what is done in `loader.CreateFromFile` function.

test/e2e/certificates_test.go
```diff
@@ -267,6 +269,69 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough clusters-65035", func() {
+
+			By("Getting Infrastructure object")
+			infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if infra.Status.PlatformStatus.Type != configv1.GCPPlatformType {
+				Skip("Skip this case for current plaform, it's not gcp.")
```
swghosh, line 279: Just to clarify, we're skipping this test if Infrastructure spec doesn't match platform type GCP, right?
Need suggestions, if this is a good approach to do :) plausibly other test examples using a similar `Skip()` would help support IMO. 

Alternatively, please check if we could use some sort of labels within ginkgo to filter out basis of platform and in our `make test-e2e` we can pass the filter arg in some way to ensure we're intending to run on AWS/GCP. IMHO, that would serve as better long-term solution as the number of test cases for specific cloud platform could grow over time and hence, it be easy to filter from the entrypoint of the tests than within the tests itself.

swghosh (commented):
Thanks @geliu2016 for the new test cases!

test/e2e/certificates_test.go
xingxingxia: @geliu2016 As said in Slack, please also update the `It()` string for the AWS scenario as above two comments. Thanks!

test/e2e/utils_test.go
geliu2016: @swghosh There is write file step in func: replaceStr()
err = ioutil.WriteFile(fileName, []byte(f1), 0644)
but it seems go embed.FS only support ReadFile, do you have any experience?

test/e2e/certificates_test.go
geliu2016: @swghosh Yes, this 'Skip' just using for filter test case doesn't match platform type GCP, 
we also think about use labels within ginkgo to filter the platfrom, but there is not good way to realize it.

test/e2e/utils_test.go
swghosh: As [embed.FS](https://pkg.go.dev/embed) is a read-only file system it doesn't support write mechanisms.
For your specific case, I see that write file is explicitly not required.

swghosh: My suggestion is something similar to:

`$ git apply modified.patch`

```diff
diff --git a/test/e2e/certificates_test.go b/test/e2e/certificates_test.go
index 440bb2c7..43df958b 100644
--- a/test/e2e/certificates_test.go
+++ b/test/e2e/certificates_test.go
@@ -6,6 +6,7 @@ package e2e
 import (
 	"context"
 	"path/filepath"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -281,7 +282,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 
 			By("Check cloud credential in cluster")
 			_, err = loader.KubeClient.CoreV1().Secrets("kube-system").Get(ctx, "gcp-credentials", metav1.GetOptions{})
-			if err != nil {			
+			if err != nil {
 				Skip("Skipping for the cluster without credential in cluster")
 			}
 
@@ -316,17 +317,20 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			defer loader.DeleteTestingNS(ns.Name)
 
 			By("Creating new certificate")
-			certFile := filepath.Join("testdata", "acme", "certificate_gcp.yaml")
-			randomString := randomStr(7)
-			err = replaceStr("RANDOM_STR", randomString, certFile)
-			Expect(err).NotTo(HaveOccurred())
-			defer func() {
-				err = replaceStr(randomString, "RANDOM_STR", certFile)
-	                        Expect(err).NotTo(HaveOccurred())
-			}()
+			loadFileAndReplaceStr := func(fileName string) ([]byte, error) {
+				bytes, err := testassets.ReadFile(fileName)
+				if err != nil {
+					return nil, err
+				}
+
+				fileContentsStr := string(bytes)
+				randomString := randomStr(7)
+				fileContentsStr = strings.ReplaceAll(fileContentsStr, "RANDOM_STR", randomString)
+				return []byte(fileContentsStr), nil
+			}
 			// The name is defined by the testdata YAML file certificate_gcp.yaml
 			certName := "cert-with-acme-dns01-clouddns-ambient"
-			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), ns.Name)
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), ns.Name)
 			By("Waiting for certificate to get ready")
 			err = waitForCertificateReadiness(ctx, certName, ns.Name)
 			Expect(err).NotTo(HaveOccurred())
diff --git a/test/e2e/utils_test.go b/test/e2e/utils_test.go
index cd696847..c0832683 100644
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"math/rand"
 	"strings"
 	"sync"
@@ -341,15 +340,3 @@ func randomStr(size int) string {
 	}
 	return s.String()
 }
-
-// replace string in file
-func replaceStr(oldStr string, newStr string, fileName string) error {
-	f, err := ioutil.ReadFile(fileName)
-	if err != nil {
-		return err
-	}	
-	f1 := strings.ReplaceAll(string(f), oldStr, newStr)
-	err = ioutil.WriteFile(fileName, []byte(f1), 0644)
-	return err
-}
-

```

swghosh:
/label docs-approved
/label px-approved
Self-adding, as this PR does not introduce any user facing changes so no doc/px change.

geliu2016:
@xingxingxia @swghosh could you please take a view? thanks

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,71 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "gcp-creds", metav1.GetOptions{})
+				if err != nil {
+					return true, nil
+				}
+				return false, nil
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Configure cert-manager to use credential, setting this credential secret name in subscription object")
+			// The name is defined cloud credential in cluster ns: kube-system
```
xingxingxia, line 303: Incorrect. This one is not defined in kube-system. Please see explanations again which were given in https://github.com/openshift/cert-manager-operator/pull/166#discussion_r1436889237 and sample comment was given in https://github.com/openshift/cert-manager-operator/pull/166#discussion_r1437348369

test/e2e/testdata/acme/clusterissuer_gcp.yaml
```diff
@@ -0,0 +1,15 @@
...
+  acme:
+    preferredChain: ""
+    privateKeySecretRef:
+      name: issuer-acme-letsencrypt
+    server: https://acme-staging-v02.api.letsencrypt.org/directory
+    solvers:
+    - dns01:
+        cloudDNS:
+          # The ID of the GCP project
+          project: openshift-qe
```
xingxingxia, line 15: Please care about trailing white blank:
```suggestion
          project: openshift-qe
```

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,71 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("Creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01-65035")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("Creating new certificate")
+			loadFileAndReplaceStr := func(fileName string) ([]byte, error) {
```
xingxingxia, line 320: As we encountered before, this usage was used several times, thus in future we might use it again. So, instead of within only one `It()`, suggest move it to be a function in utility file.

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,71 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			if err != nil {
+				Skip("Skipping for the cluster without credential in cluster")
+			}
+
+			By("Creating CredentialsRequest object")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "credentials", "credentialsrequest_gcp.yaml"), "")
+
+			By("Waiting for cloud secret to be available")
+			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
+				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "gcp-creds", metav1.GetOptions{})
```
xingxingxia, line 294: `"gcp-creds` is not right. It should be as same as `credentialSecret := "gcp-credentials"` which is defined in credentialsrequest_gcp.yaml. Your local test could not pass successfully with `"gcp-creds`. Please always **ensure** local test passed before requesting review. Thanks.

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,71 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+				if err != nil {
+					return true, nil
+				}
+				return false, nil
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Configure cert-manager to use credential, setting this credential secret name in subscription object")
+			// The name is defined cloud credential in cluster ns: kube-system
+			credentialSecret := "gcp-credentials"
```
xingxingxia, line 304: Given https://github.com/openshift/cert-manager-operator/pull/166#discussion_r1442521295 , move `credentialSecret := "gcp-credentials"` and its `//` comment up before the `err = wait.PollImmediate` line.

xingxingxia:
Done review with above 5 comments, thanks!

test/e2e/testdata/acme/clusterissuer_gcp.yaml
geliu2016: updated.

test/e2e/testdata/acme/clusterissuer_gcp.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: acme-dns01-clouddns-ambient
```
swghosh, line 4: Please add a namespace field here, any namespace value should be good eg. `namespace: hello` / `namespace: foo` etc. to avoid the test panics in dynamic_resource_loader create resource call based upon what we discussed offline.

test/e2e/testdata/acme/certificate_gcp.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Certificate
+metadata:
+  name: cert-with-acme-dns01-clouddns-ambient
+spec:
+  secretName: cert-with-acme-dns01-clouddns-ambient
+  issuerRef:
+    # The issuer created previously
+    kind: ClusterIssuer
+    name: acme-dns01-clouddns-ambient
+    dnsNames:
```
swghosh, line 11: `dnsNames` field isn't part of `spec.issuerRef`, please consider removing indent to make sure that that this becomes as `spec.dnsNames` instead!

test/e2e/testdata/acme/clusterissuer_gcp.yaml
swghosh: You could also change current dynamic_resources.go:
https://github.com/openshift/cert-manager-operator/blob/3df88ce3d01c3e9ab8d919266c4ca8bef9dc1962/test/library/dynamic_resources.go#L81-L88

into:
```go
        if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
                if overrideNamespace != "" {
			unstructuredObj.SetNamespace(overrideNamespace)
		}

		require.NotEmpty(d.t, unstructuredObj.GetNamespace(), "Namespace can not be empty!")

		
		dri = d.DynamicClient.Resource(mapping.Resource).Namespace(unstructuredObj.GetNamespace())
	} else {
```
i.e. to change the sequence, add the `if ...` before the `require.NotEmpty(...)`.

swghosh (changes requested):

geliu2016:
@xingxingxia could you please take aview? it passed succ in local.

test/e2e/utils_test.go
```diff
@@ -399,3 +401,27 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
...
+	char := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	rand.NewSource(time.Now().UnixNano())
+	var s bytes.Buffer
+	for i := 0; i < size; i++ {
+		s.WriteByte(char[rand.Int63()%int64(len(char))])
+	}
+	return s.String()
+}
+
+//replace string in bytes of ReadFile
```
xingxingxia, line 416: I remember "//xxx" without a space like "// xxx" could lead to the `ci/prow/code-check` job fail in QE private repo PRs.
```suggestion
// replace strings in bytes read from a file
```

test/e2e/utils_test.go
```diff
@@ -399,3 +401,27 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
...
+func ReplaceStrInFile(oldStr []string, newStr []string, fileName string) ([]byte, error) {
```
xingxingxia, line 417: This file's all functions are initial lower case. I do not see the necessity why this function needs initial upper case.
```suggestion
func replaceStrInFile(oldStr []string, newStr []string, fileName string) ([]byte, error) {
```

test/e2e/certificates_test.go
xingxingxia: @geliu2016 , could you better move them up as previous comment as below?
```
			// The name is defined cloud credential by the testdata YAML file.
			credentialSecret := "gcp-credentials"
			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, credentialSecret, metav1.GetOptions{})
```

xingxingxia:
Done review with three comments. Once they are updated and a pass log is gotten, I will add /lgtm.

test/e2e/utils_test.go
geliu2016: updated.

test/e2e/utils_test.go
geliu2016: updated.

geliu2016:
> Done review with three comments. Once they are updated and a pass log is gotten, I will add /lgtm.

@xingxingxia passed succ in my local after updated.
Ran 1 of 19 Specs in 179.903 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 18 Skipped

xingxingxia:
/label qe-approved
@geliu2016 please remove WIP from title.
Once ci/prow/e2e-operator new job passes, Developers can approve.

test/e2e/certificates_test.go
```diff
@@ -267,6 +269,68 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("Creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01-65035")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("Creating new certificate")
+			randomString := randomStr(7)
+			oldStr := []string{"RANDOM_STR", "NAMESPACE"}
```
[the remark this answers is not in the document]
geliu2016: @swghosh add NAMESPACE is according your suggestion the day before yesterday, and we have not set ns in cert file before, and loader.CreateFromFile step will panic, and this step passed succ after add namespace.

test/e2e/testdata/acme/certificate_gcp.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: NAMESPACE
+spec:
+  secretName: cert-with-acme-dns01-clouddns-ambient
+  issuerRef:
+    # The issuer created previously
+    kind: ClusterIssuer
+    name: acme-dns01-clouddns-ambient
+  dnsNames:
+    - RANDOM_STR.qe1.gcp.devcluster.openshift.com
+    - '*.RANDOM_STR.qe1.gcp.devcluster.openshift.com'
```
swghosh, line 14: Can you also make `qe1.gcp.devcluster.openshift.com` a placeholder like `DNS_NAME` instead.
And at runtime you could replace the same using your `replaceStrInFile` func with value of `library.GetClusterBaseDomain(ctx, configClient)`.

Something like this, inside your code of the test case:
```go
baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)

oldStr := []string{"RANDOM_STR", "DNS_NAME"}
newStr := []string{randomString, baseDomain}
// ...
```
This would ensure that when these tests are run by someone else on a non-QE infra, it still works out for them without changes!

test/e2e/utils_test.go
```diff
@@ -399,3 +401,27 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
...
+	rand.NewSource(time.Now().UnixNano())
+	var s bytes.Buffer
+	for i := 0; i < size; i++ {
+		s.WriteByte(char[rand.Int63()%int64(len(char))])
+	}
+	return s.String()
+}
+
+// replace string in bytes of ReadFile
+func replaceStrInFile(oldStr []string, newStr []string, fileName string) ([]byte, error) {
```
swghosh, line 417: Can you change `oldStr` and `newStr` arrays into a `map[string]string` instead? Seemingly that'd more accurately capture replacement logic.

eg.
`func replaceStrInFile(replacements map[string]string, fileName string) ([]byte, error)`

swghosh (changes requested):

test/e2e/utils_test.go
geliu2016: updated.

test/e2e/testdata/acme/certificate_gcp.yaml
geliu2016: updated.

geliu2016:
@swghosh pls take a view, thx, test passed succ in my local.

test/e2e/certificates_test.go
```diff
@@ -267,6 +269,72 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("Creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01-65035")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("Creating new certificate")
+			randomString := randomStr(7)
+			baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
```
swghosh, line 320: Just checked that `baseDomain` is already in `BeforeAll` section, so we can skip this line it should still work.

swghosh (approved):
Minor nit, looks good otherwise!

swghosh:
/assign @TrilokGeer 
for final approval

test/e2e/certificates_test.go
xingxingxia: @geliu2016 CPTAL ^^ ?
/hold

test/e2e/certificates_test.go
xingxingxia: @geliu2016 sorry last review did not find this is not yet been updated, CPTAL at it ^^?

test/e2e/certificates_test.go
geliu2016: updated.

xingxingxia:
/unhold
/assign @TrilokGeer 
@geliu2016 pretty awesome updates now! Reviewed new updates since my last review https://github.com/openshift/cert-manager-operator/compare/5636ad7..dfb0eed , LGTM.

Please **also** tell your new test result (just wanting to ensure PASSED) in case of any glitch. Thx!

test/e2e/testdata/acme/certificate_gcp.yaml
xingxingxia: @swghosh thanks! Actually I am aware of the need of replacing DNS_NAME. But test/e2e/testdata/acme/clusterissuer_gcp.yaml has `project: openshift-qe`.
```
        cloudDNS:
          # The ID of the GCP project
          project: openshift-qe
```
The ID of the GCP project needs to be replaced too, but seems not easy to query it in programming, so I directly just did not raise DNS_NAME replacement. @swghosh @geliu2016 @lunarwhite , any idea? Or just let as is for now?

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,70 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			By("Creating new certificate")
+			randomString := randomStr(7)
+			replaceStrMap := map[string]string{
+				"RANDOM_STR": randomString,
+				"DNS_NAME": baseDomain,
+			}
+			loadFileAndReplaceStr := func(fileName string) ([]byte, error) {
+				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
+				return []byte(fileContentsStr), err
+                        }
```
xingxingxia, line 328: @geliu2016 Please **take care** of trailing spaces at end of line.
And take care of inconsistent usage of spaces while all other lines are using \t tabs at beginning of line. Thanks
```suggestion
			}
```

lunarwhite:
Found minor places that might need update. Please give me some time to finish review, thank you!

/hold

test/e2e/certificates_test.go
geliu2016: updated.

xingxingxia:
LGTM from my side after reviewing all updates on all comments of Swarup, Yuedong and me. Pretty good now! Given Yuedong will have a few more minor comments, wait for his input.

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,70 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough clusters-65035", func() {
```
lunarwhite, line 274:
```suggestion
		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough cluster", func() {
```
In the public repo, I think it's better to not add our private case id. In future, we would find a better way to tracking/mapping cases.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
swghosh: thanks!

test/e2e/testdata/acme/certificate_gcp.yaml
swghosh: Good catch @xingxingxia, I missed it in earlier reviews.

You can use similar replacement for project ID, too.
```
infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})

gcpProjectId := infra.Status.PlatformStatus.GCP.ProjectID
Expect(gcpProjectId).NotTo(Equal(""))

// replace "GCP_PROJECT_ID" in file with the value ...
```

swghosh: As `configClient` is already used in another test case: 
```
ACME Certificate
dns-01 challenge using ambient credentials
should obtain a valid LetsEncrypt certificate
```

> The ID of the GCP project needs to be replaced too, but seems not easy to query it in programming

so guessing it should be easy to do :)

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,70 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+				return []byte(fileContentsStr), err
+			}
+			// The name is defined by the testdata YAML file certificate_gcp.yaml
+			certName := "cert-with-acme-dns01-clouddns-ambient"
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "certificate_gcp.yaml"), ns.Name)
+
+			By("Waiting for certificate to get ready")
+			err = waitForCertificateReadiness(ctx, certName, ns.Name)
+			Expect(err).NotTo(HaveOccurred())
+		})
```
lunarwhite, line 336: After `By("waiting for certificate to get ready")`, we should also add a logic to `By("checking for certificate validity from secret contents")`. Similar to https://github.com/openshift/cert-manager-operator/blob/b16d6ddca12df8143eefe78bf53e1273ebf3a8b0/test/e2e/certificates_test.go#L266-L268, and replace args with current case's.

lunarwhite (commented):
Finished from my side, thank you!

test/e2e/certificates_test.go
```diff
@@ -61,7 +63,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 	})
 
 	Context("dns-01 challenge using explicit credentials", func() {
-		It("should obtain a valid LetsEncrypt certificate", func() {
+		It("should obtain a valid LetsEncrypt certificate in AWS non-sts cluster", func() {
```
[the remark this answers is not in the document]
geliu2016: updated.

test/e2e/certificates_test.go
geliu2016: updated.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
geliu2016: updated.

test/e2e/testdata/acme/certificate_gcp.yaml
geliu2016: updated.

test/library/cert_utils.go
```diff
@@ -121,8 +122,8 @@ func VerifyCertificate(secret *v1.Secret, commonName string) (bool, error) {
 	}
 
 	// check certificate subject CN
-	if cert.Subject.CommonName != commonName {
-		return false, fmt.Errorf("incorrect subject CN: %v found in issued certificate", cert.Subject.CommonName)
+	if strings.ToLower(cert.Subject.CommonName) != strings.ToLower(commonName) {
```
swghosh, line 125: @geliu2016 Did you come across an edge case where this if condition was getting violated without the addition of `strings.ToLower`, just curious!

test/e2e/testdata/acme/clusterissuer_gcp.yaml
```diff
@@ -0,0 +1,15 @@
...
+  acme:
+    preferredChain: ""
+    privateKeySecretRef:
+      name: issuer-acme-letsencrypt
+    server: https://acme-staging-v02.api.letsencrypt.org/directory
+    solvers:
+    - dns01:
+        cloudDNS:
+          # The ID of the GCP project
+          project: OPENSHIFT-QE
```
xingxingxia, line 15: It is better to use general placeholder text instead of concrete ID value.
```suggestion
          project: PROJECT_ID
```

swghosh:
PR is in great shape, thank you for your contribution!
/lgtm

test/e2e/testdata/acme/clusterissuer_gcp.yaml
swghosh: +1

test/library/cert_utils.go
```diff
@@ -121,8 +122,8 @@ func VerifyCertificate(secret *v1.Secret, commonName string) (bool, error) {
...
+		return false, fmt.Errorf("incorrect subject CN: %v found in issued certificate, it different from %s", cert.Subject.CommonName, commonName)
```
xingxingxia, line 126: Does it fail without `strings.ToLower`?

test/library/cert_utils.go
geliu2016: Yes, 
print err log in this func as below, you may find the original random_str is: GQZqCtH, but get from func: "cert, err := GetX509Certificate(secret)", the GQZqCtH is changed to "gqzqcth", 

<*errors.errorString | 0xc00043d6b0>: 
      incorrect subject CN: gqzqcth.geliu41512.qe.gcp.devcluster.openshift.com found in issued certificate, different from GQZqCtH.geliu41512.qe.gcp.devcluster.openshift.com
      {
          s: "incorrect subject CN: gqzqcth.geliu41512.qe.gcp.devcluster.openshift.com found in issued certificate, different from GQZqCtH.geliu41512.qe.gcp.devcluster.openshift.com",
      }

swghosh: Then we should instead make sure that the random string generator function only returns lowercase characters. As DNS names are generally preferred lowercase.

swghosh: > Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical.

Excerpt from https://www.ietf.org/rfc/rfc1035.txt

test/e2e/utils_test.go
```diff
@@ -399,3 +401,27 @@ func verifyCertificateRenewed(ctx context.Context, secretName, namespace string,
 		return true, nil
 	})
 }
+
+// create randomized string
+func randomStr(size int) string {
+	char := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
xingxingxia, line 407: @geliu2016 
```suggestion
	char := "abcdefghijklmnopqrstuvwxyz0123456789"
```

test/e2e/certificates_test.go
```diff
@@ -267,6 +270,84 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "clusterissuer_gcp.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("Creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01-65035")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("Creating new certificate")
+			randomString := randomStr(7)
```
xingxingxia, line 329: @geliu2016 to avoid failure error again like OCPQE-16688, it is better to make randomString shorter so that the whole length of `randomString.baseDomain` will be shorter.
```suggestion
			randomString := randomStr(3)
```

test/e2e/testdata/acme/clusterissuer_gcp.yaml
xingxingxia: @geliu2016 it is better to use general `project: PROJECT_ID` instead of either `project: OPENSHIFT_QE` or `project: OPENSHIFT-QE`

test/e2e/utils_test.go
geliu2016: @xingxingxia both changes: char defining and add strings.ToLower are one time updating, strings.ToLower more safer than char defining, otherwise,  "aaa.Author.qe.com" still meet err compare with "aaa.author.qe.com" even define 'aaa' are small letter.

test/e2e/certificates_test.go
geliu2016: @xingxingxia ARE YOU SURE? the length of random_str is shorter at first, and you worry about that shorter randomstr will conflict with other parts of dns_name, so I added length of random_str to low the risk of same name conflict issue.
it depend on you, if you prefer shorter, I may update it.

test/e2e/testdata/acme/clusterissuer_gcp.yaml
geliu2016: updated.

test/e2e/certificates_test.go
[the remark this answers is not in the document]
geliu2016: @lunarwhite do you have any more comment? both @xingxingxia and @swghosh are giving positive comments actively, any quick response are welcome, thanks

test/e2e/utils_test.go
geliu2016: Updated.

test/e2e/certificates_test.go
geliu2016: Updated.

test/e2e/utils_test.go
xingxingxia: @geliu2016 , it is better to follow above suggestion `char := "abcdefghijklmnopqrstuvwxyz0123456789"`, it is legal to include digits within dns names.

xingxingxia:
LGTM from my side
/hold cancel
@geliu2016 once you show PASS test result, please @ our approver @TrilokGeer to /approve. Thanks!

geliu2016:
test passed succ in my local. cc: @TrilokGeer

PR #170, opened 2024-01-03
MULTIARCH-3780: Added Test Case Modification for IBMCloud and Power

pranitaT:
Issue: https://issues.redhat.com/browse/MULTIARCH-3780
updated https://github.com/openshift/cert-manager-operator/blob/master/test/e2e/certificates_test.go

Added new tests:
- dns-01 challenge using explicit credentials on ibm cloud CIS

Modified existing tests:
- dns-01 challenge using explicit credentials
- dns-01 challenge using ambient credentials

### DIFF
--- a/test/e2e/certificates_test.go
+++ b/test/e2e/certificates_test.go
@@ -5,6 +5,7 @@ package e2e
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -22,6 +23,14 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+)
+
+const (
+	// TARGET_PLATFORM is the environment variable for IBM Cloud CIS test.
+	targetPlatformEnvironmentVar = "TARGET_PLATFORM"
+
+	// CIS_CRN is the required environment variable for IBM Cloud platform.
+	cisCRNEnvironmentVar = "CIS_CRN"
 )
 
 var _ = Describe("ACME Certificate", Ordered, func() {
@@ -67,6 +76,9 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 
 	Context("dns-01 challenge using explicit credentials", func() {
 		It("should obtain a valid LetsEncrypt certificate", func() {
+			if _, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok {
+				Skip("skipping, using ibmcloud cis webhook")
+			}
 
 			By("creating a test namespace")
 			ns, err := loader.CreateTestingNS("e2e-acme-explicit-dns01")
@@ -174,10 +186,67 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		// This test uses IBM Cloud Internet Services (CIS) for the DNS-01 test cases.
+		// These tests work with both UPI / IPI installations by passing in the CRN of your CIS instance on IBM Cloud.
+		It("should obtain a valid LetsEncrypt certificate on ibm cloud CIS", func() {
+			cisCRN, isCisCRN := os.LookupEnv(cisCRNEnvironmentVar)
+			if targetPlatform, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok && targetPlatform == "ibmcloud-upi" {
+				if !isCisCRN || cisCRN == "" {
+					Fail("cisCRN is required for IBM Cloud platform")
+				}
+			} else {
+				Skip("skipping as the cluster does not use IBM Cloud CIS")
+			}
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-explicit-dns01-ibmcloud")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("creating new certificate ClusterIssuer with IBM Cloud CIS webhook solver")
+			randomString := randomStr(3)
+			clusterIssuerName := "letsencrypt-dns01-explicit-ic"
+			replaceStrMap := map[string]string{
+				"CIS_CRN": cisCRN,
+			}
+			loadFileAndReplaceStr := func(fileName string) ([]byte, error) {
+				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
+				return []byte(fileContentsStr), err
+			}
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "clusterissuer_ibmcis.yaml"), "")
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("creating new certificate")
+			// The name is defined by the testdata YAML file certificate_ibmcis.yaml
+			certDomain := "adwie." + appsDomain // acronym for "ACME dns-01 ibmcloud Webhook Explicit", short naming to pass dns name validation
+			certName := "letsencrypt-cert-ic"
+			replaceStrMap = map[string]string{
+				"RANDOM_STR": randomString,
+				"DNS_NAME":   certDomain,
+			}
+			loadFileAndReplaceStr = func(fileName string) ([]byte, error) {
+				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
+				return []byte(fileContentsStr), err
+			}
+			loader.CreateFromFile(loadFileAndReplaceStr, filepath.Join("testdata", "acme", "certificate_ibmcis.yaml"), ns.Name)
+
+			By("Waiting for certificate to get ready")
+			err = waitForCertificateReadiness(ctx, certName, ns.Name)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("checking for certificate validity from secret contents")
+			err = verifyCertificate(ctx, certName, ns.Name, randomString+"."+certDomain)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
 	})
 
 	Context("dns-01 challenge using ambient credentials", func() {
 		It("should obtain a valid LetsEncrypt certificate using ClusterIssuer on AWS mint/passthrough cluster", func() {
+			if _, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok {
+				Skip("skipping, using ibmcloud cis webhook")
+			}
 
 			By("creating a test namespace")
 			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
@@ -431,7 +500,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			randomString := randomStr(3)
 			replaceStrMap = map[string]string{
 				"RANDOM_STR": randomString,
-				"DNS_NAME": baseDomain,
+				"DNS_NAME":   baseDomain,
 			}
 			loadFileAndReplaceStr = func(fileName string) ([]byte, error) {
 				fileContentsStr, err := replaceStrInFile(replaceStrMap, fileName)
--- a/test/e2e/testdata/acme/certificate_ibmcis.yaml
+++ b/test/e2e/testdata/acme/certificate_ibmcis.yaml
@@ -1,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: letsencrypt-cert-ic
+spec:
+  dnsNames:
+    - RANDOM_STR.DNS_NAME
+    - '*.RANDOM_STR.DNS_NAME'
+  issuerRef:
+    name: letsencrypt-dns01-explicit-ic
+    kind: ClusterIssuer
+  secretName: letsencrypt-cert-ic
+  
--- a/test/e2e/testdata/acme/clusterissuer_ibmcis.yaml
+++ b/test/e2e/testdata/acme/clusterissuer_ibmcis.yaml
@@ -1,0 +1,21 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-dns01-explicit-ic
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
+    privateKeySecretRef:
+      name: letsencrypt-dns01-issuer
+    solvers:
+    - dns01:
+        webhook:
+          groupName: acme.borup.work
+          solverName: ibmcis
+          config:
+            apiKeySecretRef:
+              name: ibmcis-credentials
+              key: api-token
+            cisCRN: 
+              - "CIS_CRN"
+              


xingxingxia:
@lunarwhite @geliu2016 would you like to also help review when available?

chiragkyal:
If I'm not wrong, we don't have CI setup to run tests on IBM Power yet. Are there any plans to also add CI as follow-up PRs?
also cc: @swghosh

swghosh:
> If I'm not wrong, we don't have CI setup to run tests on IBM Power yet. Are there any plans to also add CI as follow-up PRs?

Yeah, we dunno have any plans to add CI runs for IBM P/Z systems atleast for now (we can probably look that up in future). However, these tests are intended to be skipped in the normal CI runs which we have for AWS clusters in our current jobs; so it should ideally be really similar to what is being done for GCP in #166. I believe QE intends to run these tests independently in own infra where the necessary cloud provider is available.

Nevertheless, over time we can grow test cases in this same repo and use filters to run e2e tests **(i.e. continue to use `test/e2e` as single source of all e2e cases; irrespective of what's run on pre-sumbits)** effectively per platform.

prb112:
Hey @chiragkyal 
> If I'm not wrong, we don't have CI setup to run tests on IBM Power yet. Are there any plans to also add CI as follow-up PRs?
That's right

test/e2e/certificates_test.go
```diff
@@ -435,6 +452,202 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			clusterIssuer := &certmanagerv1.ClusterIssuer{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      clusterIssuerName,
+					Namespace: ns.Name,
+				},
+				Spec: certmanagerv1.IssuerSpec{
+					IssuerConfig: certmanagerv1.IssuerConfig{
+						ACME: &v1.ACMEIssuer{
+							Server: "https://acme-v02.api.letsencrypt.org/directory",
```
lunarwhite, line 485:
```suggestion
							Server: "https://acme-staging-v02.api.letsencrypt.org/directory",
```
For e2e ci, it would be better to use [staging](https://letsencrypt.org/docs/staging-environment/) endpoint.

test/e2e/certificates_test.go
```diff
@@ -435,6 +452,202 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("checking for certificate validity from secret contents")
+			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Context("dns-01 challenge using ambient credentials on ibm cloud CIS", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
```
lunarwhite, line 555: Is this case the same as the first one? If yes, suggest to only keep the first one.

About "explicit" and "ambient", you can refer to this [doc](https://docs.openshift.com/container-platform/4.14/security/cert_manager_operator/cert-manager-operator-issuer-acme.html).

> Explicit credentials are configured by specifying the secret details in the `Issuer` API object. Ambient credentials are extracted from the environment, metadata services, or local files which are not explicitly configured in the `Issuer` API object.

test/e2e/certificates_test.go
```diff
@@ -435,6 +452,202 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
+
+	Context("dns-01 challenge using explicit credentials on ibm cloud CIS", func() {
+		It("should obtain a valid LetsEncrypt certificate", func() {
```
lunarwhite, line 457: It's better to move this case into `Context("dns-01 challenge using explicit credentials", func() { ...`

https://github.com/openshift/cert-manager-operator/blob/6833a9e979e0601713f724800be44f7b4a8792d6/test/e2e/certificates_test.go#L65

And rename the `It` description.

As below:

```
Context("dns-01 challenge using explicit credentials", func() {
    It("should obtain a valid LetsEncrypt certificate", func() {
        ...
    }
+
+   It("should obtain a valid LetsEncrypt certificate in IBM Cloud CIS", func() {
+       ...
+   }
}
```

test/e2e/certificates_test.go
```diff
@@ -435,6 +452,202 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			By("creating a test namespace")
+			ns, err := loader.CreateTestingNS("e2e-acme-explicit-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("creating new certificate ClusterIssuer with IBM Cloud CIS webhook solver")
+			clusterIssuerName := "letsencrypt-dns01-explicit"
+			By("delete existing ClusterIssuer with the same name if it exists")
+			_ = certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
```
lunarwhite, line 475: Not sure if this `Delete` operation will affect other cases (with same `clusterIssuerName`) that are running in parallel.

lunarwhite (commented):
Thanks for your work! Left some comments from my side.

test/e2e/certificates_test.go
prb112: We will update the code as such, thank you

test/e2e/certificates_test.go
prb112: It's a great question. We are following the style of the prior tests, and wrapping them where we are using a webhook to test on IBM ppc64le and s390x. Are there recommendations on updating? should we? we could create with a random hash postfix.

test/e2e/certificates_test.go
prb112: Will do

test/e2e/certificates_test.go
prb112: we are modeling the test https://github.com/openshift/cert-manager-operator/blob/2a70d97d75f2234164de73822ce43d5c63e0932d/test/e2e/certificates_test.go

We'd have to have direction to not model it, and just have one test.

prb112:
HI @lunarwhite your comments should be addressed, if acceptable, please let us know. Thank you. Paul

test/e2e/certificates_test.go
```diff
@@ -171,10 +185,109 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("should obtain a valid LetsEncrypt certificate on ibm cloud CIS", func() {
+			cisCRN, isCisCRN := os.LookupEnv(cisCRNEnvironmentVar)
+			if targetPlatform, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok && targetPlatform == "ibmcloud" {
```
swghosh, line 191: Are these values setup in IBM Cloud OpenShift clusters by default, or you're running them in a specific environment where such environment variables are available?

swghosh, line 191: Alternatively, could we resort to checking the `Infrastructure` object instead? I'm guessing that checking `infra.Status.PlatformStatus.IBMCloudPlatformStatus` is a much better option.

test/e2e/certificates_test.go
```diff
@@ -171,10 +185,109 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+							PrivateKey: certmanagermetav1.SecretKeySelector{
+								LocalObjectReference: certmanagermetav1.LocalObjectReference{
+									Name: "letsencrypt-dns01-issuer",
+								},
+							},
+							Solvers: []v1.ACMEChallengeSolver{
+								{
+									DNS01: &v1.ACMEChallengeSolverDNS01{
+										Webhook: &v1.ACMEIssuerDNS01ProviderWebhook{
+											GroupName:  "acme.borup.work",
```
swghosh, line 227: I wouldn't prefer having an explicit value here in this spec block,
what is `acme.borup.work` used for? And, is it something that is dependent on your environment or would work on all IBM Cloud clusters with CIS webhook.

test/e2e/certificates_test.go
```diff
@@ -171,10 +185,109 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("should obtain a valid LetsEncrypt certificate on ibm cloud CIS", func() {
+			cisCRN, isCisCRN := os.LookupEnv(cisCRNEnvironmentVar)
+			if targetPlatform, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok && targetPlatform == "ibmcloud" {
+				if !isCisCRN || cisCRN == "" {
```
swghosh, line 192: If I'm not mistaken, you can also get the value of `cisCRN` from `infra.Status.PlatformStatus.IBMCloudPlatformStatus.CISInstanceCRN`. Will that value work the needs of the webhook that is being introduced?

test/e2e/certificates_test.go
```diff
@@ -171,10 +185,109 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+
+			By("checking for certificate validity from secret contents")
+			err = verifyCertificate(ctx, certName, ns.Name, certDomain)
+			Expect(err).NotTo(HaveOccurred())
+		})
 	})
 
 	Context("dns-01 challenge using ambient credentials", func() {
 		It("should obtain a valid LetsEncrypt certificate in AWS mint/passthrough cluster", func() {
+			if _, ok := os.LookupEnv(targetPlatformEnvironmentVar); ok {
```
swghosh, line 288: Ditto, my other above comment. For now, let's skip if Infrastructure's platform status isn't IBM cloud. In the future, we would want to modularize these test cases so we have a specific suite for IBM Cloud, AWS, GCP, etc. etc. (today we don't have a good way of differentiation, so stuck with `Skip()`).

test/e2e/certificates_test.go
swghosh: I'd say add your own random suffix, and use the same suffix at the the time of deletion; that'd make most sense.

test/e2e/certificates_test.go
```diff
@@ -348,6 +461,102 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
+			Expect(err).NotTo(HaveOccurred())
+			defer loader.DeleteTestingNS(ns.Name)
+
+			By("creating new certificate ClusterIssuer with IBM Cloud CIS webhook solver")
+			clusterIssuerName := "letsencrypt-dns01-ambient"
+			By("delete existing ClusterIssuer with the same name if it exists")
+			_ = certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			clusterIssuer := &certmanagerv1.ClusterIssuer{
```
swghosh, line 485: You might want to avoid duplicating this ClusterIssuer block of code. Why not use a file like other test cases are doing and template out the values that needs replacement at runtime? Would also help make the entire manifest more readable.

swghosh (commented):
Thanks @pranitaT for this! I've a bunch of concerns, most of them are informational in nature since I want to understand the IBM CIS webhook integration better; and ideally want to make these test cases such that anyone with a IBM Cloud OpenShift cluster could re-run these steps with minimal setup of the webhook.

test/e2e/certificates_test.go
```diff
@@ -348,6 +461,102 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+						},
+					},
+				},
+			}
+			_, err = certmanagerClient.CertmanagerV1().ClusterIssuers().Create(ctx, clusterIssuer, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			defer certmanagerClient.CertmanagerV1().ClusterIssuers().Delete(ctx, clusterIssuerName, metav1.DeleteOptions{})
+
+			By("creating new certificate")
+			certDomain := "adwa." + appsDomain // acronym for "ACME dns-01 Webhook Ambient", short naming to pass dns name validation
```
swghosh, line 529: prefer to use a different dns name here, to avoid repetition with above.

test/e2e/certificates_test.go
```diff
@@ -348,6 +461,102 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+											SolverName: "ibmcis",
+											Config: &apiextensionsv1.JSON{
+												Raw: []byte(fmt.Sprintf(`{
+													"api-key-secret-ref": {
+														"local-object-reference": {
+															"name": "ibmcis-credentials"
+														},
+														"key": "api-token"
+													},
+													"cisCRN": ["%s"]
```
swghosh, line 513: We're explicitly passing the `cisCRN` here? Can this test case be still called "ambient credentials" use? I'm not sure.

swghosh: Instead why not refactor to use explicit credentials with a explicit auth tokens/password/file/keys in one test case, and use no token/password/file/keys in another test case where the auth mechanism of IBM CIS would be directly inferred upon from cert-manager controller pod environment implicitly!

test/e2e/certificates_test.go
prb112: makes sense.

test/e2e/certificates_test.go
prb112: This sounds like a good approach.

test/e2e/certificates_test.go
prb112: Right. Today, we use a UPI install on IBM Cloud to use with ppc64le machines.

test/e2e/certificates_test.go
prb112: It'd work with all clusters with the IBMCloud CIS Webhook. It is determined by https://github.com/IBM/cert-manager-webhook-ibmcis/blob/main/cert-manager-webhook-ibmcis.yaml#L27

test/e2e/certificates_test.go
prb112: We're running them in a specific environment where such environment variables are passed in . For example:

`TARGET_PLATFROM=ibmcloud CIS_CRN make test-e2e`

prb112: We'll double check this, however, I don't think this is going to work with the UPI install we use in IBM Cloud PowerVS.

test/e2e/certificates_test.go
prb112: This is available in the Platform object from the IPI IBMCloud VPC. It's not available to the UPI PowerVS install.

test/e2e/certificates_test.go
prb112: As a follow-up... 
```oc get infrastructure cluster -ojson | jq -r '.spec.platformSpec.type'
None
```
We cannot use this as it reports None. It is a UPI installation.

prb112: If you'd like we can add this to a ConfigMap that is manually setup.

test/e2e/certificates_test.go
prb112: Same comments... https://github.com/openshift/cert-manager-operator/pull/170/files#r1523125878

prb112:
Hey @swghosh 
Pranita has addressed the code comments, and I have responded in threads to specific points. Please let us know if you need anymore details for the review.
Thank you
Paul

pranitaT:
Hi @swghosh 
I've updated the branch. Please review the changes. 
Thank you

test/e2e/certificates_test.go
swghosh: In case that is IBM Cloud on UPI, can we make it something that can take the UPI keyword into consideration. How about "ibmcloud-upi" instead? That'd work for us for the mid term (and also helps not conflict if we have ibmcloud-ipi in the future).

swghosh: Thanks for the clarity on this, could you please help document here itself that the 
IBM CIS CRN is something that you inject from our side or if there are scripts/controller already running on your UPI cluster to do that for us.

test/e2e/certificates_test.go
swghosh: Last question, can someone run the same test if they're on a non-ppc64le IBM cloud OpenShift cluster. Probably, not right? Because PowerVS is generally handled differently from what's in s390x, amd64 IBM Cloud OpenShift clusters., ref: https://docs.openshift.com/container-platform/4.15/installing/installing_ibm_cloud_public/installing-ibm-cloud-customizations.html#installation-ibm-cloud-config-yaml_installing-ibm-cloud-customizations vs https://docs.openshift.com/container-platform/4.15/installing/installing_ibm_powervs/installing-ibm-power-vs-customizations.html#installation-ibm-power-vs-config-yaml_installing-ibm-power-vs-customizations. 

Essentially, I'm highlighting between `platform: ibmcloud` vs `platform: powervs` vs `platform: none`. Hope am not deflecting too much from the original context but my only intent here is to understand what the "Platform" view looks like from within the OpenShift cluster where you intend to run these test cases.

swghosh (approved):
This looks okay in terms of test implementation from my side, at this point. I just have asked a few inline questions to primarily clarify about the target platform (from an OpenShift purview).

Also, if you could help with a sign-off that the tests worked on your infra in it's latest state right before merge, it'd be very helfpul since there's no CI here to absolutely guarantee things went well.

test/e2e/certificates_test.go
prb112: Sure - @pranitaT can you switch the var to ibmcloud-upi

prb112: Hey @pranitaT can you add that please

test/e2e/certificates_test.go
prb112: Yes, we have one team running the tests with s390x on premises. The only part that doesn't work for them is the http01.

swghosh:
/cc @TrilokGeer 
FYI, improvements regarding modularization of test case structure for different cloud providers can be follow-ed up in CM-261.

prb112:
Hi @TrilokGeer thoughts on this code? thanks, Paul

swghosh:
@prb112 would need https://github.com/openshift/cert-manager-operator/pull/185 to land first in order for that fips-image-scan test to turn green.

prb112:
Got it. thanks @swghosh

prb112:
Hi @TrilokGeer Hope all is well. back to you for review, Paul

prb112:
Hey Swarup, do we need to rebase for this test to pass? Thanks, Paul

prb112:
Hey @TrilokGeer please take a look, thanks, Paul

swghosh:
/label docs-approved
/label px-approved
These are non-user facing changes to e2e test cases only. Multi-arch operator already landed in 1.13+.

swghosh:
/cc @lunarwhite 
for qe-approved label

prb112:
Pinged Pranita to look at conflicts

prb112:
Hey @TrilokGeer @swghosh Pranita has resolved the merge conflicts. Thank you, Paul

PR #103, opened 2023-02-13
CFE-755: Support cluster-wide egress proxy injection

thejasn:
Description
---

This PR adds egress proxy injection support to the operator. Proxy environment variables will now be propagated to the operand from the operator deployment. This PR also provides an option to provide custom trusted CA certificates. 

- `pkg/controller/deployment/cert_manager_cainjector_deployment.go`, `pkg/controller/deployment/cert_manager_controller_set.go`, `pkg/controller/deployment/cert_manager_controller_deployment.go` and `pkg/controller/deployment/cert_manager_webhook_deployment.go` : Fix informers arg names and pass trusted CA configmap at runtime. 
- `pkg/controller/deployment/deployment_overrides.go` : Add 2 new deployment hooks to inject proxy env variables and inject trusted CA configmap as a volume. 
- `pkg/controller/deployment/generic_deployment_controller.go` : Add configmap informers to re-sync deployments (watch configmaps in the operand namespace).

PR #171, opened 2024-01-03
CM-183: Add infrastructure feature annotations

swghosh:
Added Infrastructure features for OpenShift catalog metadata as per [CFC best practices](https://docs.engineering.redhat.com/display/CFC/Best_Practices).

PS: OCPBUGS-8665 reference for addition of `token-auth-azure`. While `token-auth-aws` and `token-auth-gcp` was implemented previously in https://github.com/openshift/cert-manager-operator/pull/120 and `proxy-aware` in https://github.com/openshift/cert-manager-operator/pull/103.

### DIFF
--- a/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
@@ -212,6 +212,13 @@ metadata:
     console.openshift.io/disable-operand-delete: "true"
     containerImage: openshift.io/cert-manager-operator:latest
     createdAt: 2023-03-03T00:00:00
+    features.operators.openshift.io/disconnected: "false"
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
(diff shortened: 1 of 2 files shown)


xingxingxia:
I read the `features.operators.openshift.io/cnf|cni|csi` "annotations can be omitted". So adding:
/qe-approved
Backport to older branches if needed.
FYI @lunarwhite

swghosh:
I think @xingxingxia you wanted to mean
/label qe-approved

swghosh:
/label docs-approved
(no doc req for this change immediately)

swghosh:
/cc @davemulford 
for px-approved label

swghosh:
/assign @TrilokGeer 
JFYI, the CVP tests impose a warning right now without these feature annotations but in near future the absence of these annotations would result in gated test error in our release.

xingxingxia:
```
/uncc @deads2k
```

Do we need BTW update obsolete https://github.com/openshift/cert-manager-operator/blob/master/OWNERS in separate PR?

swghosh:
> Do we need BTW update obsolete https://github.com/openshift/cert-manager-operator/blob/master/OWNERS in separate PR?

Yes, need to do that; to update that file.

PR #176, opened 2024-03-04
CM-262: Updates for Azure ambient credentials

swghosh:
- Update `docs/cloud_credentials.md` for adding Azure Workload Identity steps.
- Add ambient credentials support for Azure clusters without Workload Identity - allows cluster administrators to inject a specific (explicit) cloud credential secret for Azure clusters, similar to what we already do for AWS non-STS and GCP non-Workload Identity.

### DIFF
--- a/docs/cloud_credentials.md
+++ b/docs/cloud_credentials.md
@@ -76,15 +76,15 @@ ls <output-dir>/manifests/*-credentials.yaml | xargs -I{} oc apply -f {}
 oc delete pods --all -n cert-manager
 ```
 
-Wait for the pods to come up in running state.
-
-```sh
-oc get pods -n cert-manager -w
-```
-
 7. Patch the subscription object on the cluster to inject the secret name in the operator deployment.
 ```sh
 oc -n cert-manager-operator patch subscription <subscription-name> --type='merge' -p '{"spec":{"config":{"env":[{"name":"CLOUD_CREDENTIALS_SECRET_NAME","value":"aws-creds"}]}}}'
+```
+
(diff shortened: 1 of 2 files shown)


xingxingxia:
@swghosh I had thought about non-workload-identy ambient support after testing Azure workload identity cluster and thus I created a "Test"-only Jira CM-262. But later I thought it might be not yet supported by our code, so we move it to To Do. I'm glad to see you also think of it and give support to it. Therefore, let's met retitle:

/title CM-262: Updates for Azure ambient credentials

xingxingxia:
@swghosh I had thought about non-workload-identy ambient support after testing Azure workload identity cluster and thus I created a "Test"-only Jira [CM-262](https://issues.redhat.com//browse/CM-262). But later I thought it might be not yet supported by our code, so we move it to To Do. I'm glad to see you also think of it and give support to it. Therefore, let's met retitle:

/retitle [CM-262](https://issues.redhat.com//browse/CM-262): Updates for Azure ambient credentials

xingxingxia:
/retitle CM-262: Updates for Azure ambient credentials

lunarwhite (commented):
/label qe-approved

swghosh:
/hold
for now, I've a security concern regarding exposing the Client Secret into the environment variable directly in the deployment. This would mean for a non-Workload Identity Azure cluster is someone gets access to the Deployment yaml, they'll get access to cloud-credentials secret values and as they're long-lived credentials it might  not a good idea to do it this way.

I'll try looking for alternatives to instead use a secret mounted as a file, and still make it work, similar to what we do for AWS and GCP.

PR #179, opened 2024-04-08
CM-257: Add case to use Issuer with "--issuer-ambient-credentials" on AWS

lunarwhite:
Configuring an ACME Issuer by using ambient credentials on AWS

- Add `--issuer-ambient-credentials` argument in ACME Certificate's `BeforeAll()`
- Similair to "[should obtain a valid LetsEncrypt certificate in AWS mint/passthrough cluster](https://github.com/openshift/cert-manager-operator/blob/207d9062d78f16b786998a7989d2487eee77bf21/test/e2e/certificates_test.go#L177)", but change to use `Issuer`

### DIFF
--- a/test/e2e/certificates_test.go
+++ b/test/e2e/certificates_test.go
@@ -38,10 +38,13 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 		Expect(baseDomain).NotTo(BeEmpty())
 		appsDomain = "apps." + baseDomain
 
-		By("adding override args for dns-01 private zone passthrough")
+		By("adding override args to cert-manager controller")
 		err = addOverrideArgs(certmanageroperatorclient, certmanagerControllerDeployment, []string{
+			// for dns-01 private zone passthrough
 			"--dns01-recursive-nameservers-only",
 			"--dns01-recursive-nameservers=8.8.8.8:53,1.1.1.1:53",
+			// for Issuer to use ambient credentials
+			"--issuer-ambient-credentials",
 		})
 		Expect(err).NotTo(HaveOccurred())
 
@@ -174,7 +177,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 	})
 
 	Context("dns-01 challenge using ambient credentials", func() {
-		It("should obtain a valid LetsEncrypt certificate in AWS mint/passthrough cluster", func() {
+		It("should obtain a valid LetsEncrypt certificate using ClusterIssuer on AWS mint/passthrough cluster", func() {
 
 			By("creating a test namespace")
 			ns, err := loader.CreateTestingNS("e2e-acme-ambient-dns01")
@@ -188,9 +191,9 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
 				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "aws-creds", metav1.GetOptions{})
 				if err != nil {
-					return true, nil
+					return false, nil
 				}
-				return false, nil
+				return true, nil
 			})
 			Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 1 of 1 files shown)


lunarwhite:
Can observe the e2e-operator job succeeded: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cert-manager-operator/179/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1777625924615802880

Changing to "Ready for review", but not urgent.

/cc @swghosh @xingxingxia

test/e2e/certificates_test.go
```diff
@@ -188,9 +191,9 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
 				_, err := loader.KubeClient.CoreV1().Secrets("cert-manager").Get(ctx, "aws-creds", metav1.GetOptions{})
 				if err != nil {
-					return true, nil
+					return false, nil
 				}
-				return false, nil
+				return true, nil
```
xingxingxia, line 196: Good catch.

test/e2e/certificates_test.go
```diff
@@ -271,7 +274,106 @@ var _ = Describe("ACME Certificate", Ordered, func() {
...
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+			_, err = certmanagerClient.CertmanagerV1().Issuers(ns.Name).Create(ctx, issuer, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			defer certmanagerClient.CertmanagerV1().Issuers(ns.Name).Delete(ctx, issuerName, metav1.DeleteOptions{})
```
xingxingxia, line 340: Just note it is not needed. Issuer would be deleted when the test namespace goes to be deleted. But it is OK to delete explicitly.

test/e2e/certificates_test.go
```diff
@@ -174,7 +177,7 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 	})
 
 	Context("dns-01 challenge using ambient credentials", func() {
-		It("should obtain a valid LetsEncrypt certificate in AWS mint/passthrough cluster", func() {
+		It("should obtain a valid LetsEncrypt certificate using ClusterIssuer on AWS mint/passthrough cluster", func() {
```
xingxingxia, line 180: Given we are adding more scenarios, let's retitle so that even glance at title can get the exact checkpoint.
```suggestion
		It("should obtain a valid LetsEncrypt certificate using ambient ClusterIssuer on AWS mint/passthrough cluster", func() {
```

test/e2e/certificates_test.go
```diff
@@ -271,7 +274,106 @@ var _ = Describe("ACME Certificate", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
-		It("should obtain a valid LetsEncrypt certificate in GCP mint/passthrough clusters-65035", func() {
+		It("should obtain a valid LetsEncrypt certificate using Issuer on AWS mint/passthrough cluster", func() {
```
xingxingxia, line 277:
```suggestion
		It("should obtain a valid LetsEncrypt certificate using ambient Issuer on AWS mint/passthrough cluster", func() {
```

test/e2e/certificates_test.go
lunarwhite: Prefer to retain this step for consistency for now

test/e2e/certificates_test.go
lunarwhite: Will update

lunarwhite: About `ambient`/`explicit`, is already mentioned here:

https://github.com/openshift/cert-manager-operator/blob/207d9062d78f16b786998a7989d2487eee77bf21/test/e2e/certificates_test.go#L176

https://github.com/openshift/cert-manager-operator/blob/207d9062d78f16b786998a7989d2487eee77bf21/test/e2e/certificates_test.go#L65

swghosh (approved):
/lgtm
This looks great, sorry for the extra delays in landing this.

Non-user facing changes, only e2e test case additions:
/label px-approved
/label docs-approved

PR #185, opened 2024-05-23
CM-333: Go build(s) with strictfipsruntime

swghosh:
- `strictfipsruntime` mode for Go build
- build operand on CI (instead of mirroring from jetstack upstream)
- enable `features.operators.openshift.io/fips-compliant` in CSV

Refer https://github.com/openshift/release/pull/51700 for rehearsed PoC.

### DIFF
--- a/images/ci/Dockerfile
+++ b/images/ci/Dockerfile
@@ -1,12 +1,13 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
 WORKDIR /go/src/github.com/openshift/cert-manager-operator
+
+# build operator
 COPY . .
 RUN make build --warn-undefined-variables
 
-FROM registry.access.redhat.com/ubi9-minimal:9.2
+FROM registry.access.redhat.com/ubi9-minimal:9.4
 COPY --from=builder /go/src/github.com/openshift/cert-manager-operator/cert-manager-operator /usr/bin/
 
 USER 65532:65532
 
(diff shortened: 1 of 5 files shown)


swghosh:
/hold
until https://github.com/openshift/release/pull/51700 is merged

images/ci/Dockerfile
```diff
@@ -1,12 +1,13 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
 WORKDIR /go/src/github.com/openshift/cert-manager-operator
+
+# build operator
 COPY . .
 RUN make build --warn-undefined-variables
 
-FROM registry.access.redhat.com/ubi9-minimal:9.2
+FROM registry.access.redhat.com/ubi9-minimal:9.4
```
TrilokGeer, line 8: should the base image used to be changed to rhel els?

swghosh: I checked that the EUS content image is not available on our CI registry in an attempt to not leak the private EUS content freely outside. However, I've updated midstream to use the ELS image instead of UBI and that should use the correct base image for FIPS.

Also, check-payload tool for verifying the binaries to be FIPS compliant technically still passes with UBI base as long as dynamic linking and CGO is enabled.

swghosh:
/label docs-approved
doc labels already in place for 1.14 JIRAs
/label px-approved
self-adding as 1.14 was prior approved

lunarwhite (commented):
/label qe-approved

PR #186, opened 2024-05-28
OCPBUGS-24193: Use `imagePullPolicy: IfNotPresent` for operator deployment

swghosh:
This avoids blocking upgrades in a limited connectivity/airgapped env where on prior images are pulled before an upgrade, which is per [the related enhancement guidance](https://github.com/openshift/enhancements/blob/master/enhancements/update/dont-require-registry-during-reboot-and-upgrade.md#dont-use-the-always-pull-policy).


lunarwhite (commented):
/label qe-approved

swghosh:
/label docs-approved
added docs-needed label to OCPBUGS-24193 Jira issue for docs tracker.

PR #189, opened 2024-06-06
OCPBUGS-34698: Fix for TestCertRenew sporadic failures

bharath-b-rh:
PR contain below changes:
- Moved securityContext properties in ACME deployment to container spec.
- Added debug logs on TestCertRenew.
- Changes variable names from snake case to camel case in TestCertRenew.
- Compare certificate's pre-fetched NotAfter value from status subresource against the dynamically fetched certificate in the poll block, which guess was causing the sporadic failures.

### DIFF
--- a/test/e2e/cert_manager_deployment_test.go
+++ b/test/e2e/cert_manager_deployment_test.go
@@ -157,10 +157,10 @@ func TestCertRenew(t *testing.T) {
 	ctx := context.Background()
 	loader := library.NewDynamicResourceLoader(ctx, t)
 	config, err := library.GetConfigForTest(t)
-	require.NoError(t, err)
+	require.NoErrorf(t, err, "failed to fetch host configuration: %v", err)
 
 	ns, err := loader.CreateTestingNS("e2e-cert-renew")
-	require.NoError(t, err)
+	require.NoErrorf(t, err, "failed to create namespace: %v", err)
 	defer loader.DeleteTestingNS(ns.Name)
 	loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
 	defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
@@ -174,13 +174,13 @@ func TestCertRenew(t *testing.T) {
 	defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "acme", "service.yaml"), ns.Name)
 
 	configClient, err := configv1.NewForConfig(config)
-	require.NoError(t, err)
+	require.NoErrorf(t, err, "failed to create config client: %v", err)
 	baseDomain, err := library.GetClusterBaseDomain(ctx, configClient)
-	require.NoError(t, err)
+	require.NoErrorf(t, err, "failed to fetch cluster base domain: %v", err)
 	appsDomain := "apps." + baseDomain
 
-	ingress_host := "ecr." + appsDomain
-	path_type := networkingv1.PathTypePrefix
+	ingressHost := "ecr." + appsDomain
+	pathType := networkingv1.PathTypePrefix
 	ingress := &networkingv1.Ingress{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "networking.k8s.io/v1",
@@ -193,13 +193,13 @@ func TestCertRenew(t *testing.T) {
 		Spec: networkingv1.IngressSpec{
 			Rules: []networkingv1.IngressRule{
 				{
-					Host: ingress_host,
+					Host: ingressHost,
 					IngressRuleValue: networkingv1.IngressRuleValue{
 						HTTP: &networkingv1.HTTPIngressRuleValue{
 							Paths: []networkingv1.HTTPIngressPath{
 								{
 									Path:     "/",
-									PathType: &path_type,
+									PathType: &pathType,
 									Backend:  networkingv1.IngressBackend{Service: &networkingv1.IngressServiceBackend{Name: "hello-openshift", Port: networkingv1.ServiceBackendPort{Number: 8080}}},
 								},
 							},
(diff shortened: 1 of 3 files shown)


test/e2e/cert_manager_deployment_test.go
```diff
@@ -237,47 +237,67 @@ func TestCertRenew(t *testing.T) {
...
+			t.Errorf("failed to fetch certificate after creation: %v", err)
+			return false, nil
+		}
+		for _, condition := range crt.Status.Conditions {
+			if condition.Type == certmanagerv1.CertificateConditionReady && condition.Status == certmanagermetav1.ConditionTrue {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
```
lunarwhite, line 257: Maybe we could replace these with the utils func here: https://github.com/openshift/cert-manager-operator/blob/67689422c30cd0b487966666542df3afef791042/test/e2e/utils_test.go#L334

lunarwhite (commented):
Really appreciate your work! Just leave one minor suggestion, otherwise LGTM

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: Agree, suggestion incorporated.

bharath-b-rh: @lunarwhite ^^

lunarwhite:
/lgtm

/assign @swghosh @TrilokGeer 
for the final review/merge

test/e2e/testdata/acme/deployment.yaml
```diff
@@ -26,11 +26,11 @@ spec:
               protocol: TCP
             - containerPort: 8888
               protocol: TCP
+          securityContext:
```
TrilokGeer, line 29: What is the issue faced in having security context at pod level rather than moving to container specific?

test/e2e/cert_manager_deployment_test.go
```diff
@@ -236,48 +236,62 @@ func TestCertRenew(t *testing.T) {
...
 		},
 	}
-	certmanagerv1Client, err := certmanagerclientset.NewForConfig(config)
-	require.NoError(t, err)
-	crt, err = certmanagerv1Client.CertmanagerV1().Certificates(crt.ObjectMeta.Namespace).Create(ctx, crt, metav1.CreateOptions{})
-	defer certmanagerv1Client.CertmanagerV1().Certificates(crt.ObjectMeta.Namespace).Delete(ctx, crt.ObjectMeta.Name, metav1.DeleteOptions{})
-	require.NoError(t, err)
-	err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
+	// certmanagerClient is defined in utils_test and reusing it here for waitForCertificateReadiness().
+	certmanagerClient, err = certmanagerclientset.NewForConfig(config)
```
TrilokGeer, line 240: Is the certmanagerClient overlapping with ginkgo and go test frameworks because of using the same package?

test/e2e/cert_manager_deployment_test.go
```diff
@@ -236,48 +236,62 @@ func TestCertRenew(t *testing.T) {
...
 			t.Errorf("Expiry %v", err)
 			return false, nil
 		}
-		expiryTime, err := library.GetCertExpiry(ingress_host+":443", tlsConfig.Clone())
-		t.Logf("Expiry Before %v", expiryTime)
-		if err != nil {
-			return false, nil
-		}
+
 		time.Sleep(time.Minute + time.Second*5)
```
TrilokGeer, line 275: Optimization is required to wait for exactly the remaining time, using time.After() might help

test/e2e/cert_manager_deployment_test.go
```diff
@@ -236,48 +236,62 @@ func TestCertRenew(t *testing.T) {
...
+	defer certmanagerClient.CertmanagerV1().Certificates(crt.Namespace).Delete(ctx, crt.Name, metav1.DeleteOptions{})
+
+	err = waitForCertificateReadiness(ctx, crt.Name, crt.Namespace)
+	require.NoError(t, err, "failed while waiting for certificate to be ready: %v", err)
+
+	crt, err = certmanagerClient.CertmanagerV1().Certificates(crt.Namespace).Get(ctx, crt.Name, metav1.GetOptions{})
+	require.NoError(t, err, "failed to fetch certificate: %v", err)
+
+	// Store the NotAfter validity time of the newly created certificate.
+	expiryTime := crt.Status.NotAfter.Time.UTC()
```
TrilokGeer, line 253: Should we use `expiryTime, err := library.GetCertExpiry(ingress_host+":443", tlsConfig.Clone()) `for consistency with testcase of validating tls certificate renewal?

test/e2e/cert_manager_deployment_test.go
```diff
@@ -236,48 +236,62 @@ func TestCertRenew(t *testing.T) {
...
-	require.NoError(t, err)
-	err = wait.PollImmediate(PollInterval, TestTimeout, func() (bool, error) {
+	// certmanagerClient is defined in utils_test and reusing it here for waitForCertificateReadiness().
+	certmanagerClient, err = certmanagerclientset.NewForConfig(config)
+	require.NoErrorf(t, err, "failed to create cert manager client: %v", err)
+	_, err = certmanagerClient.CertmanagerV1().Certificates(crt.Namespace).Create(ctx, crt, metav1.CreateOptions{})
+	require.NoError(t, err, "failed to create certificate: %v", err)
+	defer certmanagerClient.CertmanagerV1().Certificates(crt.Namespace).Delete(ctx, crt.Name, metav1.DeleteOptions{})
+
+	err = waitForCertificateReadiness(ctx, crt.Name, crt.Namespace)
```
TrilokGeer, line 246: nit: add a newbie comment that says 'validates whether certificate is issued successfully'

test/e2e/testdata/acme/deployment.yaml
bharath-b-rh: Yeah, properties `allowPrivilegeEscalation`, `capabilities` are not part of pod securityContext, but container, so moved to it. We can add other properties defined for pod as well, but since just one container is defined, thought to keep just the current changes. Please let me know WDYT.

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: Yeah, `waitForCertificateReadiness` makes use of the global parameter `certmanagerClient` and expects to be initialized when invoking it.

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: I thought to use the time updated by the operator to compare it against received certificate, so that status update is also validated. Please let me know WDYT?

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: Let me check this, I don't think it's required, since polling is used.

TrilokGeer: `time.Sleep(time.Minute + time.Second*5)` needs refinement.

test/e2e/cert_manager_deployment_test.go
TrilokGeer: It'd be better to use a local variable scoped specifically than the overlapping value.

bharath-b-rh: Yeah, but that would mean refactoring the code. For now I will create a local function where a variable is passed.

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: suggestion incorporated.

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: suggestion incorporated.

test/e2e/cert_manager_deployment_test.go
bharath-b-rh: @TrilokGeer ^^

bharath-b-rh:
@TrilokGeer ^^

bharath-b-rh:
/label px-approved
/label docs-approved
no user facing changes

bharath-b-rh:
@lunarwhite Could you please add qe-approved label

PR #181, opened 2024-04-09
CM-318: Add case to use explicit credentials in ACME DNS-01 on GCP

lunarwhite:
Configuring an ACME issuer by using explicit credentials for GCP CloudDNS

Similair to "[should obtain a valid LetsEncrypt certificate](https://github.com/openshift/cert-manager-operator/blob/207d9062d78f16b786998a7989d2487eee77bf21/test/e2e/certificates_test.go#L66)", but change to use CloudDNS DNS-01 solver

Pass log:

```
go test \
-timeout 1h \
-count 1 \
-v \
-p 1 \
-tags e2e \
-run "" \
./test/e2e \
-ginkgo.label-filter="Platform: isSubsetOf {GCP}"

------------------------------
ACME Certificate
/Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:35
  dns-01 challenge with Google CloudDNS [Platform:GCP]
  /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:383
    should obtain a valid LetsEncrypt certificate using explicit credentials with ClusterIssuer
    /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:384
  > Enter [BeforeAll] ACME Certificate - /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:40 @ 11/07/24 21:40:32.073
  STEP: creating Kube clients - /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:41 @ 11/07/24 21:40:32.073
  STEP: adding override args to cert-manager controller - /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:49 @ 11/07/24 21:40:32.345
  < Exit [BeforeAll] ACME Certificate - /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:40 @ 11/07/24 21:40:32.898 (825ms)
  > Enter [BeforeEach] ACME Certificate - /Users/yuewu/Documents/workspace/fork/ocp-cert-manager-operator/test/e2e/certificates_test.go:66 @ 11/07/24 21:40:32.898
[...]

PR #194, opened 2024-07-24
CM-261: utilize ginkgo label filter to differentiate cloud platforms

lunarwhite:
Main changes:
- Bump onsi/gingko/v2 to [v2.19](https://github.com/onsi/ginkgo/releases/tag/v2.19.0).
  - To make use of the newly introduced [Label Sets](https://onsi.github.io/ginkgo/#label-sets) feature. Benefits:
    - It is more expressive and human-readable than regexp combining `--focus` and `--skip`.
    - It extends the current `--label-filter` to support filtering by key-value pairs, with the contains and not contains operations for the labels that have values. (more example usages can be found in the [doc](https://onsi.github.io/ginkgo/#label-sets))
    - It gives more flexibility in filtering and sorting tests based on labels. As the test suite grows, we might apply more label types to identify Serial/Disruptive, Alpha/Beta, Conformance/MinVersion:xxx, etc.
- Add an env variable `E2E_GINKGO_LABEL_FILTER` in the Makefile. Append `-ginkgo.label-filter=$(E2E_GINKGO_LABEL_FILTER)` to the go test command.
- Apply labels `"Platform:AWS"`, `"Platform:GCP"` and `"Platform:IBM"` to cases description accordingly.
- Set the default value of `E2E_GINKGO_LABEL_FILTER` in the Makefile to `"Platform: isSubsetOf {AWS}"`.
  - It would then select and run both non-platform-specific and AWS-only cases.

Other changes:
- Remove some hardcoded skipping logic. 
- Regroup ACME dns-01 cases by their provider platform. Each dns-01 provider now will have a dedicated `Context()`.

I've tested on the GCP cluster. It shows: (skipped 3 cases labeled `"Platform:AWS"`, 1 labeled `"Platform:IBM"`)

```
go test \
-timeout 1h \
-count 1 \
-v \
-p 1 \
-tags e2e \
-run "" \
./test/e2e \
-ginkgo.label-filter="Platform: isSubsetOf {GCP}"

...

Ran 17 of 21 Specs in 524.481 seconds
SUCCESS! -- 17 Passed | 0 Failed | 0 Pending | 4 Skipped
--- PASS: TestAll (524.49s)
PASS
ok  	github.com/openshift/cert-manager-operator/test/e2e	846.604s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,9 @@ SHORTCOMMIT ?= $(shell git rev-parse --short HEAD)
 GOBUILD_VERSION_ARGS = -ldflags "-X $(PACKAGE)/pkg/version.SHORTCOMMIT=$(SHORTCOMMIT) -X $(PACKAGE)/pkg/version.COMMIT=$(COMMIT)"
 
 E2E_TIMEOUT ?= 1h
+# E2E_GINKGO_LABEL_FILTER is ginkgo label query for selecting tests. See
+# https://onsi.github.io/ginkgo/#spec-labels. The default is to run tests on the AWS platform.
+E2E_GINKGO_LABEL_FILTER ?= "Platform: isSubsetOf {AWS}"
 
 MANIFEST_SOURCE = https://github.com/cert-manager/cert-manager/releases/download/v1.15.2/cert-manager.yaml
 
@@ -246,7 +249,8 @@ test-e2e: test-e2e-wait-for-stable-state
 	-p 1 \
 	-tags e2e \
 	-run "$(TEST)" \
-	./test/e2e
+	./test/e2e \
+	-ginkgo.label-filter=$(E2E_GINKGO_LABEL_FILTER)
 
 test-e2e-wait-for-stable-state:
 	@echo "---- Waiting for stable state ----"
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/go-jsonnet v0.17.0
 	github.com/mogensen/kubernetes-split-yaml v0.3.0
-	github.com/onsi/ginkgo/v2 v2.17.2
+	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
 	github.com/openshift/api v0.0.0-20240812094746-86145edb40cf
 	github.com/openshift/build-machinery-go v0.0.0-20240419090851-af9c868bcf52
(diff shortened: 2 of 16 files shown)


lunarwhite:
/uncc deads2k
/cc @swghosh @TrilokGeer 

CI is green. Mark it as ready for review now.

To keep it simple and easy for review and updates, I created this separate PR for CM-261. I didn't touch the main logic of the existing cases, so your comments in #181 are not fully addressed. After this is done, we can rebase and update #181.

Makefile
```diff
@@ -246,7 +249,8 @@ test-e2e: test-e2e-wait-for-stable-state
 	-p 1 \
 	-tags e2e \
 	-run "$(TEST)" \
-	./test/e2e
+	./test/e2e \
+	-ginkgo.label-filter=$(E2E_GINKGO_LABEL_FILTER)
```
swghosh, line 253: Does addition of this skip the non-ginkgo tests present in `test/e2e/cert_manager_deployment_test.go`?
It seems fine to me if it does, because those tests are essentially repetition of what's also in the ginkgo suite so we can skip them during e2e runs moving forward (we have had this behaviour tested for long and it isn't a problem).

Makefile
```diff
@@ -268,10 +272,10 @@ test-e2e-debug-cluster:
 	- oc logs deployment/cert-manager-operator -n cert-manager-operator
 	@echo "---- /Debugging the current state ----"
 .PHONY: test-e2e-debug-cluster
- 
+
 .PHONY: lint
-lint: 
-	$(GOLANGCI_LINT) run --config .golangci.yaml	
+lint:
```
swghosh, line 277: Not sure why these whitespace changes? (maybe your local linter is mis-behaving)

swghosh (approved):
Generally, LGTM
great work @lunarwhite thanks for this!
I'll take another closer look into the `test/e2e/certificates_test.go` later during the week.

Makefile
lunarwhite: This filter only applies to tests in the Ginkgo framework. The `test/e2e/cert_manager_deployment_test.go` go tests would not be affected.

Makefile
lunarwhite: Will undo these changes as they are not related to this PR. (Yeah. I've set the `"files.trimTrailingWhitespace": true` in my vscode editor)

swghosh:
/lgtm
/label docs-approved
/label px-approved
(non-user facing, only e2e test related changes)

lunarwhite:
self-adding
/label qe-approved

PR #205, opened 2024-10-09
CM-409,CM-410: Rebase v1.14.1 operator based on upstream cert-manager v1.14.7

swghosh:
[[v1.14.7](https://github.com/cert-manager/cert-manager/releases/tag/v1.14.7)] Rebase Checklist:

- [x] go dependencies
- [x] update bundle description to include upstream version of cert-manager
- [x] skipRange: ">=1.14.0 <1.14.1", replaces: "1.14.0"
- [x] downstream fork updates
    - [x] https://github.com/openshift/jetstack-cert-manager/releases/tag/v1.14.7
- [x] CI PR:
    - [x] https://github.com/openshift/release/pull/57877

PR #206, opened 2024-10-17
CM-409: Update replaces and skipRange to include 1.14.1

swghosh:
As a result of the 1.14.1 z-stream, we need to update the OLM replaces and skipRange to factor in the v1.14.1 inclusion for the update graph.

This a follow-up over #205.


lunarwhite (commented):
/label qe-approved

PR #221, opened 2024-12-09
CM-436: Cleanup kube-rbac-proxy sidecar

swghosh:
Fixes https://github.com/openshift/cert-manager-operator-release/issues/95

Authorization & Authentication already exists in https://github.com/openshift/cert-manager-operator/blob/master/vendor/github.com/openshift/library-go/pkg/controller/controllercmd/builder.go#L312-L337.


swghosh:
/label px-approved
No ui/ux changes

bharath-b-rh:
nit: label `app.kubernetes.io/component: kube-rbac-proxy` exists in below manifests, could be cleaned up too.
- config/rbac/auth_proxy_role.yaml
- config/rbac/auth_proxy_client_clusterrole.yaml

/lgtm

swghosh:
/label docs-approved 
We'll get it added in release notes for 1.15.0 and 1.14.2 (cc: @subhtk) that redundant kube-rbac-proxy sidecar have been removed.

swghosh:
/label qe-approved
In theory, QE already approves of this change: self-adding the label since @lunarwhite is OOO.

PR #220, opened 2024-11-27
CM-418,CM-419: Adds new istio-csr CRD and controller to manage lifecycle for the CR

bharath-b-rh:
PR implements the changes proposed in [enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/cert-manager/istio-csr-controller.md), which are
- Add a new CRD for user to indicate the preferences for istio-csr agent deployment.
- Add static manifests for deploying istio-csr, which will be updated by the controller based on the user configuration.
- Add controller to manage istio-csr deployment to meet the user configuration.

Observations:
- Leader election fails during upgrade waiting on the lease to be given up by the current instance and the new instance exits waiting for the lease. Also since `liveness` and `readiness` checks are added [here](https://github.com/openshift/cert-manager-operator/blob/14e73cf0c403c722f3c506baac67377f2f628317/config/manager/manager.yaml#L116-L129), the current instance was not stopped until the new instance was ready, which happens only when it becomes the leader. And this was causing the new instance to be in `CrashLoopBackoff`, and workaround was to delete the current instance. Hence the checks were removed.

Later, when both `cert-manager` and `istio-csr` controllers make use of `controller-runtime` lib, leader election need to enabled [here](https://github.com/openshift/cert-manager-operator/blob/14e73cf0c403c722f3c506baac67377f2f628317/pkg/operator/setup_manager.go#L53-L104), along with the health and ready checks.

PR #226, opened 2024-12-26
CM-423: Adds e2e test for istio-csr controller with grpc CreateCertificate call

PillaiManish:
The PR adds the e2e test scenarios for the new istio-csr controller.

- Apply's the new IstioCSR resource in the istio-system namespace
- Create a new job - grpcurl-job which calls the endpoint of the istio-csr
- Waits for the response and validates the certificate and the root chain

### DIFF
--- a/test/e2e/istio_csr_test.go
+++ b/test/e2e/istio_csr_test.go
@@ -1,0 +1,287 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/url"
+	"path/filepath"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
+
+	"github.com/openshift/cert-manager-operator/test/library"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// backOffLimit is the max retries for the Job
+const backOffLimit int32 = 10
+
+// istioCSRProtoURL links to proto for istio-csr API spec
+const istioCSRProtoURL = "https://raw.githubusercontent.com/istio/api/v1.24.1/security/v1alpha1/ca.proto"
+
+type LogEntry struct {
+	CertChain []string `json:"certChain"`
+}
+
+var _ = Describe("Istio-CSR", Ordered, Label("TechPreview", "Feature:IstioCSR"), func() {
+	ctx := context.TODO()
+	var clientset *kubernetes.Clientset
+	var dynamicClient *dynamic.DynamicClient
+
+	BeforeAll(func() {
+		var err error
+		clientset, err = kubernetes.NewForConfig(cfg)
+		Expect(err).Should(BeNil())
+
+		dynamicClient, err = dynamic.NewForConfig(cfg)
+		Expect(err).Should(BeNil())
+	})
+
+	var ns *corev1.Namespace
+
+	BeforeEach(func() {
+		By("waiting for operator status to become available")
+		err := verifyOperatorStatusCondition(certmanageroperatorclient, []string{
+			certManagerControllerDeploymentControllerName,
+			certManagerWebhookDeploymentControllerName,
+			certManagerCAInjectorDeploymentControllerName,
+		}, validOperatorStatusConditions)
+		Expect(err).NotTo(HaveOccurred(), "Operator is expected to be available")
+
+		By("creating a test namespace")
+		namespace, err := loader.CreateTestingNS("istio-system", true)
+		Expect(err).NotTo(HaveOccurred())
+		ns = namespace
+
+		DeferCleanup(func() {
+			loader.DeleteTestingNS(ns.Name, func() bool { return CurrentSpecReport().Failed() })
+		})
+	})
+
+	Context("grpc call istio.v1.auth.IstioCertificateService/CreateCertificate to istio-csr agent", func() {
+		It("should return cert-chain as response", func() {
+			serviceAccountName := "cert-manager-istio-csr"
+			grpcAppName := "grpcurl"
+
+			By("creating cluster issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+
+			By("issuing TLS certificate")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+
+			By("fetching proto file from api")
+			protoContent, err := library.FetchFileFromURL(istioCSRProtoURL)
+			Expect(err).Should(BeNil())
+			Expect(protoContent).NotTo(BeEmpty())
+
+			By("creating proto config map")
+			configMap := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "proto-cm",
+					Namespace: ns.Name,
+				},
+				Data: map[string]string{
+					"ca.proto": protoContent,
+				},
+			}
+			_, err = clientset.CoreV1().ConfigMaps(ns.Name).Create(ctx, configMap, metav1.CreateOptions{})
+			Expect(err).Should(BeNil())
+			defer clientset.CoreV1().ConfigMaps(ns.Name).Delete(ctx, configMap.Name, metav1.DeleteOptions{})
+
+			By("creating istio-ca issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_ca_issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_ca_issuer.yaml"), ns.Name)
+
+			By("creating istiocsr.operator.openshift.io resource")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_csr.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_csr.yaml"), ns.Name)
+
+			By("poll till cert-manager-istio-csr is available")
+			err = pollTillDeploymentAvailable(ctx, clientset, ns.Name, "cert-manager-istio-csr")
+			Expect(err).Should(BeNil())
+
+			istioCSRGRPCEndpoint, err := pollTillIstioCSRAvailable(ctx, dynamicClient, ns.Name, "default")
+			Expect(err).Should(BeNil())
+
+			By("poll till the service account is available")
+			err = pollTillServiceAccountAvailable(ctx, clientset, ns.Name, serviceAccountName)
+			Expect(err).Should(BeNil())
+
+			By("generate csr request")
+
+			csrTemplate := &x509.CertificateRequest{
+				Subject: pkix.Name{
+					Organization:       []string{"My Organization"},
+					OrganizationalUnit: []string{"IT Department"},
+					Country:            []string{"US"},
+					Locality:           []string{"Los Angeles"},
+					Province:           []string{"California"},
+				},
+				URIs: []*url.URL{
+					{Scheme: "spiffe", Host: "cluster.local", Path: "/ns/istio-system/sa/cert-manager-istio-csr"},
+				},
+				SignatureAlgorithm: x509.SHA256WithRSA,
+			}
+
+			csr, err := library.GenerateCSR(csrTemplate)
+			Expect(err).Should(BeNil())
+
+			By("creating an grpcurl job")
+			job := &batchv1.Job{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "grpcurl-job",
+				},
+				Spec: batchv1.JobSpec{
+					Completions:  ptr.To(int32(1)),
+					BackoffLimit: ptr.To(backOffLimit),
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: grpcAppName,
+							Labels: map[string]string{
+								"app": grpcAppName,
+							},
+						},
+						Spec: corev1.PodSpec{
+							ServiceAccountName:           serviceAccountName,
+							AutomountServiceAccountToken: ptr.To(false),
+							RestartPolicy:                corev1.RestartPolicyOnFailure,
+							Containers: []corev1.Container{
+								{
+									Name:  grpcAppName,
+									Image: "registry.redhat.io/rhel9/go-toolset",
+									Command: []string{
+										"/bin/sh",
+										"-c",
+									},
+									Env: []corev1.EnvVar{
+										{
+											Name:  "GOCACHE",
+											Value: "/tmp/go-cache",
+										},
+										{
+											Name:  "GOPATH",
+											Value: "/tmp/go",
+										},
+									},
+									Args: []string{
+										"go install github.com/fullstorydev/grpcurl/cmd/grpcurl@v1.9.2 >/dev/null 2>&1 && " +
+											"TOKEN=$(cat /var/run/secrets/istio-ca/token) && " +
+											"/tmp/go/bin/grpcurl " +
+											"-import-path /proto " +
+											"-proto /proto/ca.proto " +
+											"-H \"Authorization: Bearer $TOKEN\" " +
+											fmt.Sprintf("-d '{\"csr\": \"%s\", \"validity_duration\": 3600}' ", csr) +
+											"-cacert /etc/root-secret/ca.crt " +
+											"-key /etc/root-secret/tls.key " +
+											"-cert /etc/root-secret/tls.crt " +
+											fmt.Sprintf("%s istio.v1.auth.IstioCertificateService/CreateCertificate", istioCSRGRPCEndpoint),
+									},
+									VolumeMounts: []corev1.VolumeMount{
+										{Name: "root-secret", MountPath: "/etc/root-secret"},
+										{Name: "proto", MountPath: "/proto"},
+										{Name: "sa-token", MountPath: "/var/run/secrets/istio-ca"},
+									},
+								},
+							},
+							Volumes: []corev1.Volume{
+								{
+									Name: "sa-token",
+									VolumeSource: corev1.VolumeSource{
+										Projected: &corev1.ProjectedVolumeSource{
+											DefaultMode: ptr.To(int32(420)),
+											Sources: []corev1.VolumeProjection{
+												{
+													ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+														Audience:          "istio-ca",
+														ExpirationSeconds: ptr.To(int64(3600)),
+														Path:              "token",
+													},
+												},
+											},
+										},
+									},
+								},
+								{
+									Name: "root-secret",
+									VolumeSource: corev1.VolumeSource{
+										Secret: &corev1.SecretVolumeSource{
+											SecretName: "istiod-tls",
+										},
+									},
+								},
+								{
+									Name: "proto",
+									VolumeSource: corev1.VolumeSource{
+										ConfigMap: &corev1.ConfigMapVolumeSource{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "proto-cm",
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+			_, err = clientset.BatchV1().Jobs(ns.Name).Create(context.TODO(), job, metav1.CreateOptions{})
+			Expect(err).Should(BeNil())
+			defer clientset.BatchV1().Jobs(ns.Name).Delete(ctx, job.Name, metav1.DeleteOptions{})
+
+			By("waiting for the job to be completed")
+			err = pollTillJobCompleted(ctx, clientset, ns.Name, "grpcurl-job")
+			Expect(err).Should(BeNil())
+
+			By("fetching logs of the grpcurl job")
+			pods, err := clientset.CoreV1().Pods(ns.Name).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("app=%s", grpcAppName),
+			})
+			Expect(err).Should(BeNil())
+
+			By("fetching succeeded pod name")
+			var succeededPodName string
+			for _, pod := range pods.Items {
+				if pod.Status.Phase == corev1.PodSucceeded {
+					succeededPodName = pod.Name
+				}
+			}
+			Expect(succeededPodName).ShouldNot(BeEmpty())
+
+			req := clientset.CoreV1().Pods(ns.Name).GetLogs(succeededPodName, &corev1.PodLogOptions{})
+			logs, err := req.Stream(context.TODO())
+			Expect(err).Should(BeNil())
+
+			defer logs.Close()
+
+			logData, err := io.ReadAll(logs)
+			Expect(err).Should(BeNil())
+
+			var entry LogEntry
+			err = json.Unmarshal(logData, &entry)
+			Expect(err).Should(BeNil())
+			Expect(entry.CertChain).ShouldNot(BeEmpty())
+
+			By("validating each certificate")
+			for _, certPEM := range entry.CertChain {
+				err = library.ValidateCertificate(certPEM, "my-selfsigned-ca")
+				Expect(err).Should(BeNil())
+			}
+
+		})
+	})
+})
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"math/rand"
 	"regexp"
 	"strings"
@@ -25,6 +26,8 @@ import (
 	certmanoperatorclient "github.com/openshift/cert-manager-operator/pkg/operator/clientset/versioned"
 	"github.com/openshift/cert-manager-operator/test/library"
 
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -35,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
 )
@@ -562,4 +566,131 @@ func waitForIngressReadiness(ctx context.Context, client kubernetes.Interface, i
 		}
 		return false, nil
 	})
+}
+
+// pollTillJobCompleted poll the job object and returns non-nil error
+// once the job is completed, otherwise should return a time-out error
+func pollTillJobCompleted(ctx context.Context, clientset *kubernetes.Clientset, namespace, jobName string) error {
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		job, err := clientset.BatchV1().Jobs(namespace).Get(ctx, jobName, metav1.GetOptions{})
+
+		if err != nil {
+			return false, err
+		}
+
+		for _, cond := range job.Status.Conditions {
+			if cond.Type == batchv1.JobComplete {
+				if cond.Status == corev1.ConditionTrue {
+					return true, nil
+				} else {
+					return false, nil
+				}
+			}
+		}
+
+		return false, nil
+	})
+	return err
+}
+
+// pollTillServiceAccountAvailable poll the service account object and returns non-nil error
+// once the service account is available, otherwise should return a time-out error
+func pollTillServiceAccountAvailable(ctx context.Context, clientset *kubernetes.Clientset, namespace, serviceAccountName string) error {
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		_, err := clientset.CoreV1().ServiceAccounts(namespace).Get(ctx, serviceAccountName, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+
+		return true, nil
+	})
+
+	return err
+}
+
+// pollTillIstioCSRAvailable poll the istioCSR object and returns non-nil error and istio-grpc-endpoint
+// once the istiocsr is available, otherwise should return a time-out error
+func pollTillIstioCSRAvailable(ctx context.Context, dynamicClient *dynamic.DynamicClient, namespace, istioCsrName string) (string, error) {
+	var istioCSRGRPCEndpoint string
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		gvr := schema.GroupVersionResource{
+			Group:    "operator.openshift.io",
+			Version:  "v1alpha1",
+			Resource: "istiocsrs",
+		}
+
+		customResource, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(ctx, istioCsrName, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		status, found, err := unstructured.NestedMap(customResource.Object, "status")
+		if err != nil {
+			return false, nil
+		}
+
+		if !found {
+			return false, nil
+		}
+
+		conditions, found, err := unstructured.NestedSlice(customResource.Object, "status", "conditions")
+		if err != nil {
+			return false, nil
+		}
+
+		if !found {
+			return false, nil
+		}
+
+		for _, condition := range conditions {
+			condMap, ok := condition.(map[string]interface{})
+			if !ok {
+				continue
+			}
+
+			condType, _ := condMap["type"].(string)
+			condStatus, _ := condMap["status"].(string)
+
+			if condType != "Ready" {
+				continue
+			}
+
+			if condStatus == string(metav1.ConditionTrue) {
+				break
+			} else {
+				return false, nil
+			}
+
+		}
+
+		if !library.IsEmptyString(status["istioCSRGRPCEndpoint"]) && !library.IsEmptyString(status["clusterRoleBinding"]) && !library.IsEmptyString(status["istioCSRImage"]) && !library.IsEmptyString(status["serviceAccount"]) {
+			istioCSRGRPCEndpoint = status["istioCSRGRPCEndpoint"].(string)
+			return true, nil
+		}
+		return false, nil
+	})
+
+	return istioCSRGRPCEndpoint, err
+}
+
+func pollTillDeploymentAvailable(ctx context.Context, clientSet *kubernetes.Clientset, namespace, deploymentName string) error {
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		deployment, err := clientSet.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		for _, cond := range deployment.Status.Conditions {
+			if cond.Type == appsv1.DeploymentAvailable {
+				return cond.Status == corev1.ConditionTrue, nil
+			}
+		}
+
+		return false, nil
+	})
+
+	return err
 }
--- a/test/library/utils.go
+++ b/test/library/utils.go
@@ -5,8 +5,15 @@ package library
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
+	"io"
 	"log"
+	"net/http"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -17,15 +24,20 @@ import (
 	configv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 )
 
-func (d DynamicResourceLoader) CreateTestingNS(namespacePrefix string) (*corev1.Namespace, error) {
+func (d DynamicResourceLoader) CreateTestingNS(namespacePrefix string, noSuffix bool) (*corev1.Namespace, error) {
 	namespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: fmt.Sprintf("%v-", namespacePrefix),
 			Labels: map[string]string{
 				"e2e-test": "true",
 				"operator": "openshift-cert-manager-operator",
 			},
 		},
+	}
+
+	if noSuffix {
+		namespace.ObjectMeta.Name = namespacePrefix
+	} else {
+		namespace.ObjectMeta.GenerateName = fmt.Sprintf("%v-", namespacePrefix)
 	}
 
 	var got *corev1.Namespace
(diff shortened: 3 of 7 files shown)


swghosh:
/testwith openshift/cert-manager-operator/master/e2e-operator openshift/cert-manager-operator#220

bharath-b-rh:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,383 @@
...
+	loader := library.NewDynamicResourceLoader(ctx, t)
+
+	// create test namespace and deploy test components
+	testNS, err := loader.CreateNS("test")
+	require.NoErrorf(t, err, "failed to create test namespace")
+	defer loader.DeleteTestingNS(testNS.Name, t.Failed)
+
+	//csv check
+	// deploy istio service mesh and check the operator status
+	loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-subscription.yaml"), "openshift-operators")
```
TrilokGeer, line 152: Which version is being deployed for service mesh?

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,383 @@
...
+					headers[strings.TrimSpace(parts[0])] = strings.TrimSpace(parts[1])
+				}
+			}
+
+			return stdout, stderr, headers, statusCode, nil
+		}
+
+		// Log and wait before retrying
+		fmt.Printf("Retrying (%d/3)...\n", attempt)
+		time.Sleep(2 * time.Second)
```
TrilokGeer, line 114: It is better to avoid hard coding, use scoped variables or constants enables a single point of change control and better code management.

test/e2e/istio_csr_test.go
PillaiManish: For service mesh - v2.6.4 is being used.

test/e2e/istio_csr_test.go
PillaiManish: Okay, will fix that.

test/library/utils.go
```diff
@@ -17,6 +17,27 @@ import (
...
+	createdNamespace, err := d.KubeClient.CoreV1().Namespaces().Create(context.TODO(), namespaceObject, metav1.CreateOptions{})
+	if err != nil {
+		return namespaceObject, err
+	}
+
+	return createdNamespace, nil
+
+}
+
 func (d DynamicResourceLoader) CreateTestingNS(namespacePrefix string) (*v1.Namespace, error) {
```
swghosh, line 41:
```suggestion
func (d DynamicResourceLoader) CreateTestingNS(namespacePrefix string, shouldGenerateSuffix bool) (*v1.Namespace, error) {
```
consider re-using this function instead of adding another `CreateNS` with necessary `true` or `false` to the later argument.

swghosh (changes requested):
Please refactor the tests using a fresh Ginkgo suite that can have a label for `TechPreview:IstioCSR` or the alike.

Also, per discussions with team earlier we can strip the service mesh functionality for now and only make this test, really test the grpc server exposed by istio-csr with a plain client.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+	})
+
+	Context("testing grpc call of istio-csr", Label("Platform: IstioCSR"), func() {
+
+		It("should return cert-chain as response", func() {
+			By("issuing cluster issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+
+			By("issuing certificate for TLS")
```
swghosh, line 114:
```suggestion
			By("issuing TLS certificate")
```

test/library/utils.go
PillaiManish: Fixed and added Ginkgo.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+
+			By("creating proto config map")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "proto-cm.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "proto-cm.yaml"), ns.Name)
+
+			By("applying istio issuer")
```
swghosh, line 122:
```suggestion
			By("creating istio-ca issuer")
```

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-issuer.yaml"), ns.Name)
+
+			By("applying istio-csr resource")
```
swghosh, line 126:
```suggestion
			By("creating istiocsr.operator.openshift.io resource")
```

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+							Labels: map[string]string{
+								"app": grpcAppName,
+							},
+						},
+						Spec: corev1.PodSpec{
+							RestartPolicy: corev1.RestartPolicyNever,
+							Containers: []corev1.Container{
+								{
+									Name:  grpcAppName,
+									Image: "fullstorydev/grpcurl:latest", // Use your custom image if applicable
```
swghosh, line 169:
```suggestion
									Image: "fullstorydev/grpcurl:latest",
```

test/e2e/testdata/istio/istio-issuer.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: istio-ca
+  namespace: istio-system
+spec:
+  ca:
+    secretName: root-secret
```
swghosh, line 8:
```suggestion
    secretName: istio-ca
```

If this Issuer intends to use the self-signed CA cert from the other `istio-certificate.yaml`, we need to make the `secretName` reflect correctly.

test/e2e/testdata/istio/istio-issuer.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: cert-manager.io/v1
```
swghosh, line 1: to avoid confusion, can we renamed this file to `istio-ca-issuer.yaml`

test/e2e/testdata/istio/istio-certificate.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: cert-manager.io/v1
```
swghosh, line 1: hmm, similarly rename this to `istio-ca-certificate.yaml`.

test/e2e/testdata/istio/istio-certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Certificate
+metadata:
+  name: istio-ca
+  namespace: istio-system
+spec:
+  isCA: true
```
swghosh, line 7:
```suggestion
  selfSigned: {}
  isCA: true
```

test/e2e/testdata/istio/istio-certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+  duration: 21600h
+  secretName: istio-ca
+  commonName: istio-ca.my-company.net
```
swghosh, line 10:
```suggestion
  commonName: istio-ca.cluster.local
```

test/e2e/testdata/istio/istio-certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+  subject:
+    organizations:
+      - my-company.net
```
swghosh, line 13:
```suggestion
      - cluster.local
```

test/library/utils.go
swghosh: per my previous comment, please look into adding a `noSuffix` arg to the existing function and reusing it, rather than having to add another new func for the same/similar purpose.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+									},
+								},
+							},
+							Volumes: []corev1.Volume{
+								{
+									Name: "cert-volume",
+									VolumeSource: corev1.VolumeSource{
+										ConfigMap: &corev1.ConfigMapVolumeSource{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "istio-ca-root-cert",
```
swghosh, line 191: does our istio-csr controller add this ConfigMap?
I couldn't spot any resource for `istio-ca-root-cert` in the `istio-system` namespace. Do you need the service mesh installed + istiod running for this (by any chance)?

Doesn't seem the same as whatever you added in the issuer or certificate yaml(s), so am worried :)

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+					Expect(err).Should(BeNil())
+
+					Expect(cert.Issuer.CommonName).Should(Equal("my-selfsigned-ca"))
+
+					// Validate the certificate's validity period
+					By("validating the certificate validity period")
+					now := time.Now()
+					Expect(cert.NotBefore).Should(BeTemporally("<=", now), "Certificate's NotBefore should be in the past or present")
+					Expect(cert.NotAfter).Should(BeTemporally(">", now), "Certificate's NotAfter should be in the future")
+
```
swghosh, line 253: use a util function that has your validations, and call it here.

swghosh: also I don't really think your `commonName` is correct?? where was it added in the CSR to verify.

swghosh: would like to see another validation for cert chain belonging to the right CA.
fine with the other validations for commonName, expiry; apart from keeping them, mainly need to validate as the cert belongs to the CA chain (eg. https://gist.github.com/devtdeng/4f6adcb5a306f2ae035a2e7d9f724d17)

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+				}
+			}
+		})
+	})
+})
+
+func pollTillCompleted(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelSelector string) error {
```
swghosh, line 260: no need to list the pods and check their status,
just poll the Job object itself and look for it's status to see completion status.

test/e2e/testdata/istio/proto-cm.yaml
```diff
@@ -0,0 +1,65 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: proto-cm
+  namespace: istio-system
+immutable: false
+data:
+  ca.proto: |
```
swghosh, line 8: can we add a `curl <URL>` for the ca.proto directly inside the Job?

The `<URL>` can be from raw.githubusercontent.com link, **stick the test case to a known good istio api tag** from https://github.com/istio/api/tags (basis of whatever we use in https://github.com/openshift/cert-manager-istio-csr/blob/da02a0a7ce68c0c3fb348a517bec5e68493f31b3/go.mod#L22).

That way we have the expectation during each rebase that if necessary this istio API tag needs to be updated, helps us backtrack if at all the api breaks ever.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+		Type:  "CERTIFICATE REQUEST",
+		Bytes: csrBytes,
+	})
+
+	escapedCSR := strings.ReplaceAll(string(csrPEM), "\n", "\\n")
+
+	return escapedCSR, nil
+}
+
+var _ = Describe("Istio-CSR", Ordered, func() {
```
swghosh, line 74:
```suggestion
var _ = Describe("Istio-CSR", Ordered, Label("TechPreview", "Feature:IstioCSR"), func() {
```
ti's needed for https://github.com/openshift/release/pull/60386

want to trigger the test only when we want to (and `!TechPreview` was not used)...

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+	ctx := context.Background()
+	var clientset *kubernetes.Clientset
+	serviceAccountName := "cert-manager-istio-csr"
+	grpcAppName := "grpcurl"
```
swghosh, line 78: doesn't look relevant here, can you see if it can be moved inside the relevant block?

swghosh (changes requested):
overall the test case(s) seem to be in a good shape.

swghosh:
/testwith openshift/cert-manager-operator/master/e2e-operator openshift/cert-manager-operator#220

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"net/url"
+	"path/filepath"
+	"strings"
+	"time"
+)
```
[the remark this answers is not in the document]
PillaiManish: Updated.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+type LogEntry struct {
+	CertChain []string `json:"certChain"`
+}
+
+func generateCSR() (string, error) {
+	// Generate a private key
```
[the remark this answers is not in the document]
PillaiManish: Updated.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate private key: %w", err)
+	}
+
+	// Define the CSR template
```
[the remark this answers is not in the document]
PillaiManish: Updated.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+		SignatureAlgorithm: x509.SHA256WithRSA,
+	}
+
+	// Generate the CSR
+	csrBytes, err := x509.CreateCertificateRequest(rand.Reader, csrTemplate, privateKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to create CSR: %w", err)
+	}
+
+	// Encode the CSR to PEM format
```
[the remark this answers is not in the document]
PillaiManish: Updated.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+		DeferCleanup(func() {
+			loader.DeleteTestingNS(ns.Name, func() bool { return CurrentSpecReport().Failed() })
+		})
+	})
+
+	Context("testing grpc call of istio-csr", Label("Platform: IstioCSR"), func() {
+
+		It("should return cert-chain as response", func() {
+			By("issuing cluster issuer")
```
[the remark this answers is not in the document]
PillaiManish: Updated.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+		namespace, err := loader.CreateNS("istio-system")
+		Expect(err).NotTo(HaveOccurred())
+		ns = namespace
+
+		DeferCleanup(func() {
+			loader.DeleteTestingNS(ns.Name, func() bool { return CurrentSpecReport().Failed() })
+		})
+	})
+
+	Context("testing grpc call of istio-csr", Label("Platform: IstioCSR"), func() {
```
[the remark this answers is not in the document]
PillaiManish: Thanks! Got it and fixed.

test/e2e/istio-csr_test.go
PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Added the mentioned labels.

test/e2e/testdata/istio/istio-issuer.yaml
PillaiManish: Renamed the file.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+		return "", fmt.Errorf("failed to create CSR: %w", err)
+	}
+
+	// Encode the CSR to PEM format
+	csrPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE REQUEST",
+		Bytes: csrBytes,
+	})
+
+	escapedCSR := strings.ReplaceAll(string(csrPEM), "\n", "\\n")
```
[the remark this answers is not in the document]
PillaiManish: It was required to make the JSON payload valid. Else, it was throwing an exception for the new line characters '/n'.

test/e2e/istio-csr_test.go
PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Updated

test/e2e/testdata/istio/istio-certificate.yaml
PillaiManish: Updated.

test/e2e/testdata/istio/istio-certificate.yaml
PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Fixed.

test/e2e/istio-csr_test.go
PillaiManish: Added function in utility for validations.

test/e2e/istio-csr_test.go
swghosh: Nvm, my bad. The istio-csr agent itself copies that config map to each namespace on the cluster. But I think we could poll for that config map to be created first, the Job's Pod stay in ContainerCreating state for a long time until that config map is made available for mounting as a volume.

test/e2e/istio-csr_test.go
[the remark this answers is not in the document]
swghosh: din't mean to suggest to remove this line of code,
what I meant was to add a comment  why this is being done,
it's fine even without.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+			By("creating jwt token for header authorization")
+			tokenRequest := &authenticationv1.TokenRequest{
+				Spec: authenticationv1.TokenRequestSpec{
+					Audiences: []string{
+						"istio-ca",
+					},
+				},
+			}
+			tokenResponse, err := clientset.CoreV1().ServiceAccounts("istio-system").CreateToken(ctx, serviceAccountName, tokenRequest, metav1.CreateOptions{})
```
swghosh, line 142: we should remove this call from here, it hardcodes the token into the deployment yaml as a result of fetching it previously + we risk the chance that the token being used inside the pod can be expired.

Instead, can use: `pod.spec.volumes.projected.sources.serviceAccountToken.audience` with "istio-ca" and mount a volume on the Job's Pod spec which should take care of this JWT. Also, for it to work the grpcurl pod should use `serviceAccountName: cert-manager-istio-csr`.

swghosh: ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#tokenrequest-api

test/e2e/testdata/istio/istio-certificate.yaml
PillaiManish: This file was redundant, and is deleted now.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+								"app": grpcAppName,
+							},
+						},
+						Spec: corev1.PodSpec{
+							RestartPolicy: corev1.RestartPolicyNever,
+							Containers: []corev1.Container{
+								{
+									Name:  grpcAppName,
+									Image: "fullstorydev/grpcurl:latest", // Use your custom image if applicable
+									Args: []string{
```
bharath-b-rh, line 170: Which certificate is used for gRPC client? I think it would be better to make use of certificate created by the controller for istio. Certificate is created in configured istio's namespace, with the name `istiod`. And required dnsNames to be added to certificate can also be configured.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+										"-import-path", "/proto",
+										"-proto", "/proto/ca.proto",
+										"-H", fmt.Sprintf("Authorization: Bearer %s", tokenResponse.Status.Token),
+										"-d", fmt.Sprintf(`{"csr": "%s", "validity_duration": 3600}`, csr),
+										"-cacert", "/etc/certs/root-cert.pem",
+										fmt.Sprintf("cert-manager-istio-csr.%s.svc:443", ns.Name),
```
bharath-b-rh, line 176: Please use `status.istioCSRGRPCEndpoint` from `istiocsr` resource, which will help in validating correct value is set.

bharath-b-rh:
Please add the description.

test/e2e/testdata/istio/proto-cm.yaml
PillaiManish: Added the same to fetch the proto file directly from istio/api repo.

test/e2e/istio-csr_test.go
PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Yes, fixed the code and using the istiod-tls for certificate & private key.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+			By("applying istio issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-issuer.yaml"), ns.Name)
+
+			By("applying istio-csr resource")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-csr.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio-csr.yaml"), ns.Name)
+
+			By("poll till the service account is available")
```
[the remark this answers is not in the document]
PillaiManish: Updated according to the comment.

test/e2e/testdata/istio/istio-certificate.yaml
PillaiManish: This file is now deleted.

test/e2e/testdata/istio/istio-certificate.yaml
PillaiManish: This file is now deleted.

test/e2e/istio-csr_test.go
PillaiManish: Updated.

test/e2e/testdata/istio/istio-issuer.yaml
PillaiManish: The issuer  will be using the certificate signed by the cluster-issuer.

PillaiManish:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

PillaiManish:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

bharath-b-rh:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

swghosh:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

test/e2e/utils_test.go
```diff
@@ -564,3 +568,130 @@ func waitForIngressReadiness(ctx context.Context, client kubernetes.Interface, i
...
+				continue
+			}
+
+			if condStatus == string(metav1.ConditionTrue) {
+				break
+			} else {
+				return false, nil
+			}
+
+		}
```
swghosh, line 668: please parse the status.conditions into a metav1.Condition, and use the https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#FindStatusCondition helper func instead.

test/e2e/utils_test.go
```diff
@@ -564,3 +568,130 @@ func waitForIngressReadiness(ctx context.Context, client kubernetes.Interface, i
...
+func pollTillIstioCSRAvailable(ctx context.Context, dynamicClient *dynamic.DynamicClient, namespace, istioCsrName string) (string, error) {
+	var istioCSRGRPCEndpoint string
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		gvr := schema.GroupVersionResource{
+			Group:    "operator.openshift.io",
+			Version:  "v1alpha1",
+			Resource: "istiocsrs",
+		}
+
+		customResource, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(ctx, istioCsrName, metav1.GetOptions{})
```
swghosh, line 626: while not completely opposing use of the dynamic resource client,
see if it is convenient to use a istiocsr client directly from `"github.com/cert-manager/cert-manager/pkg/client/clientset/versioned"`

eg.
https://github.com/openshift/cert-manager-operator/blob/6fbe83159f4b7b5f886462c0b198242b4c3b1711/test/e2e/utils_test.go#L19

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
+//go:build e2e
```
swghosh, line 1: please rename this file to `istio_csr_test.go`.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
...
+			By("creating cluster issuer")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+
+			By("issuing TLS certificate")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+
+			By("fetching proto file from api")
+			istioURL := "https://raw.githubusercontent.com/istio/api/v1.24.1/security/v1alpha1/ca.proto"
```
swghosh, line 79: use as a `const` for this near the top of this file.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
...
+									Args: []string{
+										"GOCACHE=/tmp/go-cache && " +
+											"export GOPATH=/tmp/go && " +
+											"go install github.com/fullstorydev/grpcurl/cmd/grpcurl@v1.9.2 >/dev/null 2>&1 && " +
+											"TOKEN=$(cat /var/run/secrets/istio-ca/token) && " +
+											"/tmp/go/bin/grpcurl " +
+											"-import-path /proto " +
+											"-proto /proto/ca.proto " +
+											"-H \"Authorization: Bearer $TOKEN\" " +
+											fmt.Sprintf("-d '{\"csr\": \"%s\", \"validity_duration\": 3600}' ", csr) +
```
swghosh, line 160: nit(improves readability): store this in a new variable above, then use it here.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
...
+											"-cacert /etc/root-secret/ca.crt " +
+											"-key /etc/root-secret/tls.key " +
+											"-cert /etc/root-secret/tls.crt " +
+											fmt.Sprintf("%s istio.v1.auth.IstioCertificateService/CreateCertificate", istioCSRGRPCEndpoint),
```
swghosh, line 164: nit(improves readability): store this in a another new variable above, then use it here.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+				var entry LogEntry
+				err = json.Unmarshal(logData, &entry)
+				Expect(err).Should(BeNil())
+
+				By("validating each certificate")
+				for _, certPEM := range entry.CertChain {
+					err = library.ValidateCertificate(certPEM, "my-selfsigned-ca")
+					Expect(err).Should(BeNil())
+				}
```
swghosh, line 247: since these are Pods from a Job, under some circumstance we might see a pod that had failed, the Job would retry in that case.

Hence, make this loop such that we can get a "atleast once" guarantee. Also, ignore errors otherwise.

test/e2e/istio-csr_test.go
```diff
@@ -0,0 +1,252 @@
...
+			err = pollTillJobCompleted(ctx, clientset, ns.Name, "grpcurl-job")
+			Expect(err).Should(BeNil())
+
+			By("fetching logs of the grpcurl job")
+			pods, err := clientset.CoreV1().Pods(ns.Name).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("app=%s", grpcAppName),
+			})
+			Expect(err).Should(BeNil())
+
+			By("reading logs of the grpcurl job")
```
swghosh, line 228:
```suggestion
```

redundant.

PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Updated.

test/e2e/istio-csr_test.go
PillaiManish: Fixed.

test/e2e/istio-csr_test.go
PillaiManish: For retry, there should be retry attempt or time limit?

swghosh: set `.spec.completions = 1` and `.spec.backoffLimit = N`, N can be a reasonable retry limit like 10. 

https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures

swghosh: Filter out the pod loop by looking only into pods with status Completed.


what I mean is:
```

desiredPodName := ""
for _, pod := range pods.Items {
    if pod.Status.Phase == "Succeeded" {
            desiredPodName = pod.Name
    }
}

Expect(desiredPodName).ShouldNot(BeEmpty())


req := clientset.CoreV1().Pods(ns.Name).GetLogs(desiredPodName, &corev1.PodLogOptions{})
// ...
```

PillaiManish: Done.

test/e2e/istio-csr_test.go
PillaiManish: Updated

swghosh:
/label docs-approved
/label px-approved
No user facing changes, this is essentially a follow-up over #220.

bharath-b-rh:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

bharath-b-rh:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

lunarwhite:
/testwith openshift/cert-manager-operator/master/e2e-operator #220

lunarwhite:
/testwith openshift/cert-manager-operator/master/e2e-operator https://github.com/openshift/cert-manager-operator/pull/220

test/library/utils.go
```diff
@@ -88,3 +101,88 @@ func GetClusterBaseDomain(ctx context.Context, configClient configv1.ConfigV1Int
...
+			OrganizationalUnit: []string{"IT Department"},
+			Country:            []string{"US"},
+			Locality:           []string{"Los Angeles"},
+			Province:           []string{"California"},
+		},
+		URIs: []*url.URL{
+			{Scheme: "spiffe", Host: "cluster.local", Path: "/ns/istio-system/sa/cert-manager-istio-csr"},
+		},
+		SignatureAlgorithm: x509.SHA256WithRSA,
+	}
```
lunarwhite, line 173: Could we decouple the hardcoded template from this func?

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/cert-manager-operator/test/library"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const BACK_OFF_LIMITS int32 = 10
+const ISTIO_CSR_URL = "https://raw.githubusercontent.com/istio/api/v1.24.1/security/v1alpha1/ca.proto"
```
lunarwhite, line 27: It would be great if we could add a comment for each constant.

test/e2e/utils_test.go
```diff
@@ -564,3 +568,130 @@ func waitForIngressReadiness(ctx context.Context, client kubernetes.Interface, i
...
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		sa, err := clientset.CoreV1().ServiceAccounts(namespace).Get(ctx, serviceAccountName, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+
+		return sa != nil, nil
```
lunarwhite, line 609:
```suggestion
		return true, nil
```

PillaiManish: Updated.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,262 @@
...
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "cluster_issuer.yaml"), ns.Name)
+
+			By("issuing TLS certificate")
+			loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+			defer loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), ns.Name)
+
+			By("fetching proto file from api")
+			protoContent, err := library.FetchFileFromURL(ISTIO_CSR_URL)
+			Expect(err).Should(BeNil())
```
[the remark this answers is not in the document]
PillaiManish: Updated

test/e2e/istio_csr_test.go
PillaiManish: Done.

test/e2e/utils_test.go
```diff
@@ -564,3 +568,130 @@ func waitForIngressReadiness(ctx context.Context, client kubernetes.Interface, i
...
+		return false, nil
+	})
+	return err
+}
+
+// pollTillServiceAccountAvailable poll the service account object and returns non-nil error
+// once the service account is available, otherwise should return a time-out error
+func pollTillServiceAccountAvailable(ctx context.Context, clientset *kubernetes.Clientset, namespace, serviceAccountName string) error {
+	err := wait.PollUntilContextTimeout(ctx, PollInterval, TestTimeout, true, func(ctx context.Context) (bool, error) {
+		sa, err := clientset.CoreV1().ServiceAccounts(namespace).Get(ctx, serviceAccountName, metav1.GetOptions{})
```
swghosh, line 601:
```suggestion
		_, err := clientset.CoreV1().ServiceAccounts(namespace).Get(ctx, serviceAccountName, metav1.GetOptions{})
```

unused variable,
needed to make sure the e2e can build and run.

PillaiManish: Thanks, fixed.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,284 @@
...
+												Name: "proto-cm",
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			}
```
swghosh, line 239: prefer a file for this instead, as most of the field values are fixed. `replaceStrInFile` can help;
however follow-up PR is fine.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// back_off_limits is for setting max retries in the job spec
+const back_off_limits int32 = 10
+
+// istio_csr_url is for fetching proto content
+const istio_csr_url = "https://raw.githubusercontent.com/istio/api/v1.24.1/security/v1alpha1/ca.proto"
```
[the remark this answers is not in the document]
PillaiManish: Fixed.

test/e2e/istio_csr_test.go
PillaiManish: Maybe I can do it in a follow-up, if that works?

swghosh: let's de-clutter it, how about a `test/e2e/testdata/istio_csr/grpcurl_job.yaml`?

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: grpcurl-istio-csr
spec:
  backoffLimit: 10
  completions: 1
  template:
    metadata:
      labels:
        app: grpcurl-istio-csr
      name: grpcurl-istio-csr
    spec:
      automountServiceAccountToken: false
      containers:
      - args:
        - |
          go install github.com/fullstorydev/grpcurl/cmd/grpcurl@v1.9.2 >/dev/null 2>&1 && \
          TOKEN=$(cat /var/run/secrets/istio-ca/token) && \
          /tmp/go/bin/grpcurl \
            -import-path /proto \
            -proto /proto/ca.proto \
            -H "Authorization: Bearer $TOKEN" \
            -d '{"csr": "{{.CertificateSigningRequest}}", "validity_duration": 3600}' \
            -cacert /etc/root-secret/ca.crt \
            -key /etc/root-secret/tls.key \
            -cert /etc/root-secret/tls.crt \
            {{.IstioCSRStatus.IstioCSRGRPCEndpoint}} istio.v1.auth.IstioCertificateService/CreateCertificate
        command:
        - /bin/sh
        - -c
        env:
        - name: GOCACHE
          value: /tmp/go-cache
        - name: GOPATH
          value: /tmp/go
        image: registry.redhat.io/rhel9/go-toolset
        name: grpcurl
        volumeMounts:
        - mountPath: /etc/root-secret
          name: root-secret
        - mountPath: /proto
          name: proto
        - mountPath: /var/run/secrets/istio-ca
          name: sa-token
      restartPolicy: OnFailure
      serviceAccountName: '{{.IstioCSRStatus.ServiceAccount}}'
      volumes:
      - name: sa-token
        projected:
          defaultMode: 420
          sources:
          - serviceAccountToken:
              audience: istio-ca
              expirationSeconds: 3600
              path: token
      - name: root-secret
        secret:
          secretName: istiod-tls
      - configMap:
          name: proto-cm
        name: proto
```

what I meant basically pushes ahead of https://github.com/openshift/cert-manager-operator/pull/247, so that within `test/e2e/config_template.go` we can add:
```diff
+ // IstioCSRConfig customizes the fields in a job spec
+ type IstioCSRGRPCurlJobConfig struct {
+ 	CertificateSigningRequest string
+ 	IstioCSRStatus            v1alpha1.IstioCSRStatus
+ }
```
to get the above yaml value filled in during test.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,284 @@
...
+									VolumeMounts: []corev1.VolumeMount{
+										{Name: "root-secret", MountPath: "/etc/root-secret"},
+										{Name: "proto", MountPath: "/proto"},
+										{Name: "service-token", MountPath: "/var/run/secrets/istio-ca"},
+									},
+								},
+							},
+							Volumes: []corev1.Volume{
+								{
+									Name: "service-token",
```
swghosh, line 201:
```suggestion
									Name: "sa-token",
```

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,284 @@
...
+									},
+									Env: []corev1.EnvVar{
+										{
+											Name:  "GOCACHE",
+											Value: "/tmp/go-cache",
+										},
+									},
+									Args: []string{
+										"GOCACHE=/tmp/go-cache && " +
+											"export GOPATH=/tmp/go && " +
```
swghosh, line 179: please move `GOCACHE` and `GOPATH` into the env stanza instead.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,284 @@
...
+											fmt.Sprintf("-d '{\"csr\": \"%s\", \"validity_duration\": 3600}' ", csr) +
+											"-cacert /etc/root-secret/ca.crt " +
+											"-key /etc/root-secret/tls.key " +
+											"-cert /etc/root-secret/tls.crt " +
+											fmt.Sprintf("%s istio.v1.auth.IstioCertificateService/CreateCertificate", istioCSRGRPCEndpoint),
+									},
+									VolumeMounts: []corev1.VolumeMount{
+										{Name: "root-secret", MountPath: "/etc/root-secret"},
+										{Name: "proto", MountPath: "/proto"},
+										{Name: "service-token", MountPath: "/var/run/secrets/istio-ca"},
```
swghosh, line 195:
```suggestion
										{Name: "sa-token", MountPath: "/var/run/secrets/istio-ca"},
```

test/e2e/testdata/istio/istio-ca-issuer.yaml
swghosh, line 1: rename: `istio_ca_issuer.yaml`

test/e2e/testdata/istio/istio-csr.yaml
swghosh, line 1: rename: `istio_csr.yaml`

swghosh (approved):
few nits, otherwise LGTM.

test/e2e/istio_csr_test.go
swghosh: As discussed, we'll do this on a follow-up PR after #247 lands.

test/e2e/istio_csr_test.go
PillaiManish: Updated.

test/e2e/testdata/istio/istio-ca-issuer.yaml
PillaiManish: Fixed.

test/e2e/testdata/istio/istio-csr.yaml
PillaiManish: Fixed.

test/e2e/istio_csr_test.go
PillaiManish: Fixed.

swghosh:
/retitle CM-423: Adds e2e test for istio-csr controller with grpc CreateCertificate call

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,286 @@
...
+			Expect(err).Should(BeNil())
+
+			defer logs.Close()
+
+			logData, err := io.ReadAll(logs)
+			Expect(err).Should(BeNil())
+
+			var entry LogEntry
+			err = json.Unmarshal(logData, &entry)
+			Expect(err).Should(BeNil())
```
lunarwhite, line 276:
```suggestion
			Expect(err).Should(BeNil())
			Expect(entry.CertChain).ShouldNot(BeEmpty())
```
adding a checkpoint to ensure the chain is not empty, otherwise the for loop won't execute

lunarwhite (commented):
LGTM, only minor nits

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,286 @@
...
+			By("fetching logs of the grpcurl job")
+			pods, err := clientset.CoreV1().Pods(ns.Name).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("app=%s", grpcAppName),
+			})
+			Expect(err).Should(BeNil())
+
+			By("fetching succeeded pod name")
+			var succeededPodName string
+			for _, pod := range pods.Items {
+				if pod.Status.Phase == "Succeeded" {
```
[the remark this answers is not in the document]
PillaiManish: Fixed.

test/e2e/istio_csr_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+// istioCSRProtoURL links to proto for istio-csr API spec
+const istioCSRProtoURL = "https://raw.githubusercontent.com/istio/api/v1.24.1/security/v1alpha1/ca.proto"
+
+type LogEntry struct {
+	CertChain []string `json:"certChain"`
+}
+
+var _ = Describe("Istio-CSR", Ordered, Label("TechPreview", "Feature:IstioCSR"), func() {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
PillaiManish: Fixed

test/e2e/istio_csr_test.go
PillaiManish: Updated

PR #229, opened 2025-01-08
CM-424: Allow experimental features to be turned on from `--unsupported-addon-features`

swghosh:
Adds `--unsupported-addon-features` arg to the operator binary, which when contains `"IstioCSR=true"` turns on the controller-runtime based istio-csr controller (which was added in #220).

From a user perspective, if they wish to use unsupported addons they can patch the OLM Subscription object with the `UNSUPPORTED_ADDON_FEATURES` env variable for it to get enabled at operator startup.

### DIFF
--- a/api/operator/v1alpha1/features.go
+++ b/api/operator/v1alpha1/features.go
@@ -1,0 +1,22 @@
+package v1alpha1
+
+import (
+	"k8s.io/component-base/featuregate"
+)
+
+var (
+	// TechPreview: v1.15
+	//
+	// IstioCSR enables the controller for istiocsr.operator.openshift.io resource,
+	// which extends cert-manager-operator to deploy and manage the istio-csr agent.
+	// OpenShift Service Mesh facilitates the integration and istio-csr is an agent that
+	// allows Istio workload and control plane components to be secured using cert-manager.
+	//
+	// For more details,
+	// https://github.com/openshift/enhancements/blob/master/enhancements/cert-manager/istio-csr-controller.md
+	FeatureIstioCSR featuregate.Feature = "IstioCSR"
+)
+
+var OperatorFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	FeatureIstioCSR: {Default: false, PreRelease: "TechPreview"},
+}
--- a/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
@@ -673,6 +673,7 @@ spec:
                 - --v=$(OPERATOR_LOG_LEVEL)
                 - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
                 - --cloud-credentials-secret=$(CLOUD_CREDENTIALS_SECRET_NAME)
+                - --unsupported-addon-features=$(UNSUPPORTED_ADDON_FEATURES)
                 command:
                 - /usr/bin/cert-manager-operator
                 env:
@@ -706,6 +707,7 @@ spec:
                   value: "2"
                 - name: TRUSTED_CA_CONFIGMAP_NAME
                 - name: CLOUD_CREDENTIALS_SECRET_NAME
+                - name: UNSUPPORTED_ADDON_FEATURES
                 image: openshift.io/cert-manager-operator:latest
                 imagePullPolicy: IfNotPresent
                 name: cert-manager-operator
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -62,6 +62,7 @@ spec:
             - '--v=$(OPERATOR_LOG_LEVEL)'
             - '--trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)'
             - '--cloud-credentials-secret=$(CLOUD_CREDENTIALS_SECRET_NAME)'
+            - '--unsupported-addon-features=$(UNSUPPORTED_ADDON_FEATURES)'
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
@@ -93,6 +94,7 @@ spec:
               value: '2'
             - name: TRUSTED_CA_CONFIGMAP_NAME
             - name: CLOUD_CREDENTIALS_SECRET_NAME
+            - name: UNSUPPORTED_ADDON_FEATURES
           image: controller:latest
           imagePullPolicy: IfNotPresent
           name: cert-manager-operator
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -20,5 +20,16 @@ func NewOperator() *cobra.Command {
 	cmd.Short = "Start the cert-manager Operator"
 	cmd.Flags().StringVar(&operator.TrustedCAConfigMapName, "trusted-ca-configmap", "", "The name of the config map containing TLS CA(s) which should be trusted by the controller's containers. PEM encoded file under \"ca-bundle.crt\" key is expected.")
 	cmd.Flags().StringVar(&operator.CloudCredentialSecret, "cloud-credentials-secret", "", "The name of the secret containing cloud credentials for authenticating using cert-manager ambient credentials mode.")
+
+	cmd.Flags().StringVar(&operator.UnsupportedAddonFeatures, "unsupported-addon-features", "",
+		`List of unsupported addon features that the operator optionally enables.
+		
+eg. --unsupported-addon-features="IstioCSR=true"
+		
+Note: Technology Preview features are not supported with Red Hat production service level agreements (SLAs)
+and might not be functionally complete. Red Hat does not recommend using them in production.
+
+These features provide early access to upcoming product features,
+enabling customers to test functionality and provide feedback during the development process.`)
 	return cmd
 }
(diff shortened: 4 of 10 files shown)


api/operator/v1alpha1/certmanager_types.go
```diff
@@ -120,6 +128,30 @@ type CertManagerScheduling struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.all(e, self.filter(x, x == e).size() == 1)",message="tech preview features must not contain duplicate values."
+	// +kubebuilder:validation:Required
+	// +required
+	TechPreview TechPreview `json:"techPreview"`
+}
+
+// TechPreview contains the list of enabled experimental features.
+// +listType=atomic
+// +kubebuilder:validation:MaxItems=64
+type TechPreview []FeatureName
```
bharath-b-rh, line 153: nit: Wouldn't a struct with feature names as boolean type members be simple and CEL validations too would be less complex?
```
type Features struct {
   IstioCSR bool
}
```

pkg/controller/certmanager/controller.go
```diff
@@ -61,26 +84,54 @@ type CertManagerReconciler struct {
...
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the CertManager object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+// Reconcile is run each time the CertManager object has a create/update/delete event.
+// All objects other than certmanager.operator.openshift.io/cluster lead to no-op syncs.
 func (r *CertManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
bharath-b-rh, line 89: nit: We could add couple of UTs here I think.

pkg/controller/certmanager/tp_block_upgrades.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	"github.com/operator-framework/operator-lib/conditions"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	techPreviewNoUpgradeReason  = "techPreviewFeaturesUpgradeRestricted"
+	techPreviewNoUpgradeMessage = "The operator installed with TechPreview features cannot be upgraded."
```
bharath-b-rh, line 13: nit: Should we add SLA disclaimer here too?

pkg/controller/certmanager/tp_block_upgrades.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil
+	}
+
+	// upgrades were already blocked, do nothing
+	if cond.Message == techPreviewNoUpgradeMessage && cond.Reason == techPreviewNoUpgradeReason &&
+		cond.Status == metav1.ConditionFalse {
+		return nil
+	}
+
+	err = upgradeCond.Set(ctx, metav1.ConditionFalse,
```
bharath-b-rh, line 33: nit

```suggestion
	return upgradeCond.Set(ctx, metav1.ConditionFalse,
```

pkg/operator/feature_controller_set.go
```diff
@@ -0,0 +1,70 @@
...
+func (fc *FeatureControllerSet) pollAndRunIfEnabled(ctx context.Context, mgr ctrl.Manager, feat cm.FeatureAccessor) {
+	shouldEnable := false
+
+	for !shouldEnable {
+		select {
+		case <-ctx.Done():
+			fc.log.V(1).Info("graceful shutdown")
+			return
+
+		case <-time.Tick(retryInterval):
```
bharath-b-rh, line 49: not a blocker, instead of periodic poll, could we share a channel between `Reconcile` and   `RunWithManagerOnceEnabled` methods, say a new field in `CertManagerReconciler` and as arg to the later method, and wait on the channel for updates?

api/operator/v1alpha1/certmanager_types.go
swghosh: yes, initially I came up with that but the CEL validation(s) in Kubernetes seem to not allow validating the way I did. There's no way to do a general "check all fields" operation as a part of Kubernetes' security model where validations need to be always explicit.

The array option allows for simpler checking without needing to be explicit.

swghosh: CEL validation like:  `oldSelf.all(field, has(self[field]))` doesn't work, which ultimately lets the user unset a feature after it has been enabled.

Also, it seemed more tedious to write an expression that does equivalent of the aforesaid: 
```cel
!has(oldSelf) || (
    has(oldSelf.istioCSR) ? has(self.istioCSR) : true &&
    has(oldSelf.newFeature1) ? has(self.newFeature1) : true &&
    has(oldSelf.newFeature2) ? has(self.newFeature2) : true
    // ...
)
```
for each and every feature we add.

pkg/operator/feature_controller_set.go
swghosh: Added an event watcher to the FeatureAccessor, which takes care of doing the similar. Events are sent by the feature accessor when a new feature is enabled, this is subscribed by the `RunWithManagerOnceEnabled` method which triggers enablement accordingly.

swghosh:
Updated the implementation for the FeatureAccessor to wrap with an `k8s.io/component-base/featuregate` 's MutableFeatureGate. This provides us the additional functionality for us to promote a TechPreview feature to GA in a new release, i.e. flip the switch between a feauture being disabled by default -> enabled by default from a single point of control.

pkg/controller/certmanager/controller.go
swghosh: Added few test cases for the basic functionality, now in `tp_status_test.go`.

pkg/operator/features/feature_accessor.go
```diff
@@ -0,0 +1,23 @@
+package features
```
TrilokGeer, line 1: Can you share the usecase for which the feature_accessor or the controller is required to monitor techpreview features?

swghosh: As discussed, we'll move to setting the addon features from OLM Subscription object so the operator's environment variable can consume it.

bharath-b-rh (commented):
/lgtm

swghosh:
/label docs-approved
https://github.com/openshift/openshift-docs/pull/89608#discussion_r1988596787

api/operator/v1alpha1/features.go
```diff
@@ -0,0 +1,22 @@
...
+	// OpenShift Service Mesh facilitates the integration and istio-csr is an agent that
+	// allows Istio workload and control plane components to be secured using cert-manager.
+	//
+	// For more details,
+	// https://github.com/openshift/enhancements/blob/master/enhancements/cert-manager/istio-csr-controller.md
+	FeatureIstioCSR featuregate.Feature = "IstioCSR"
+)
+
+var OperatorFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	FeatureIstioCSR: {Default: false, PreRelease: "TechPreview"},
```
TrilokGeer, line 21: Is there a check specific to PreRelease field?

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -673,6 +673,7 @@ spec:
                 - --v=$(OPERATOR_LOG_LEVEL)
                 - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
                 - --cloud-credentials-secret=$(CLOUD_CREDENTIALS_SECRET_NAME)
+                - --addon-features=$(ADDON_FEATURES)
```
TrilokGeer, line 676: How does this distinguish between unsupported vs supported add-ons?

api/operator/v1alpha1/features.go
swghosh: Upstream uses GA, beta, alpha for this prerelease field values;
but avoided Beta/Alpha since we do not use them in our lifecycle.

swghosh: > Is there a check specific to PreRelease field?

Added a test case in features_test to ensure that `FeatureSpec{Prerelease: "TechPreview"}` features are always disabled by default.

bundle/manifests/cert-manager-operator.clusterserviceversion.yaml
swghosh: As discussed, updated arg name to `--unsupported-addon-features`,
and additionally in operator docs the tech preview related warning will be presented to users who wish to enable IstioCSR addon feature.

api/operator/v1alpha1/features.go
lunarwhite: UX question: How would a user enable a supported addon (once goes GA), or is it enabled by default?

swghosh: Will be enabled by default when the feature moves to GA. We just need to update to `FeatureSpec{Default: true, Prerelease: "GA"}` and it should work.

lunarwhite: Well, wouldn't it be kind of expensive to enable that controller for users who don't need this addon at all? Not sure how much this would impact the overall performance. Perhaps we could experiment with it when going GA, at least it shouldn't be an issue at this time 😄

PR #259, opened 2025-03-31
ci: migrate golangci to v2 schema and correct issues

sebrandon1:
Similar to other `v2` migrations:
- https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2875

There were a number of things that the latest version of golangci-lint flagged for fixing.

Also, I added a `.github/workflows/pre-main.yaml` that contains a job to run the linter prior to merging PRs.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -1,0 +1,27 @@
+name: Test Incoming Changes
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    env:
+      SHELL: /bin/bash
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Run GolangCI-lint
+        uses: golangci/golangci-lint-action@v7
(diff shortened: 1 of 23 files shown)


sebrandon1:
It seems like Actions aren't enabled on this repository.  Is that something that can be enabled by an admin?

.github/workflows/pre-main.yaml
```diff
@@ -0,0 +1,27 @@
+name: Test Incoming Changes
```
swghosh, line 1: We could add a `make` target in this repository that runs the linter with the proposed `.golangci.yaml` config.

swghosh: And, add a new linter test onto [our CI configs](https://github.com/openshift/release/tree/master/ci-operator/config/openshift/cert-manager-operator.), feel free to raise PRs on o/release for this.

That way we do not need to introduce extraneous GitHub Actions (and end up paying for [runners](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions)).

swghosh (changes requested):
Like the idea of linting our codebase,
but we cannot use GitHub actions immediately.

.github/workflows/pre-main.yaml
sebrandon1: This PR uses `ubuntu-latest` runners which are the free tier, no billing plan required.

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

As for the make target, I agree we could also add that.  The action `golangci/golangci-lint-action@v7` automatically detects the .golangci.yml and uses those settings and manages the installation and running of the linter for us.

PR #260, opened 2025-04-01
Use `k8s.io/utils/ptr` instead of deprecated `k8s.io/utils/pointer`

sebrandon1:
Broken out from #259 into a smaller PR.

Similar to: https://github.com/openshift-kni/cnf-features-deploy/pull/2345


swghosh:
/retitle Use `k8s.io/utils/ptr` instead of deprecated `k8s.io/utils/pointer`

PR #261, opened 2025-04-01
pollimmediate deprecation: use PollUntilContextTimeout instead

sebrandon1:
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate

`PollImmediate` is deprecated and they recommend using `PollUntilContextTimeout`.

Broken out from #259 for easier reviewing.


swghosh:
Thanks @sebrandon1,
This is already being done in #257

swghosh:
/close
in favour of #257 (already added a similar commit with the same changes).

PR #268, opened 2025-04-30
CM-553: Rebase for operator v1.16.0 release with upstream cert-manager v1.16.4

siddhibhor-56:
Rebases downstream operator with upstream cert-manager v1.16.4 for next release [[v1.16.4](https://github.com/cert-manager/cert-manager/releases/tag/v1.15.5)] :

- go dependencies
- update bundle description to include upstream version of cert-manager
- skipRange: ">=1.15.1 <1.16.0", replaces: "v1.15.1"
- downstream fork updates 
-https://github.com/openshift/jetstack-cert-manager/releases/tag/v1.16.4
- Follow up PR Over https://github.com/openshift/cert-manager-operator/pull/263

PR #272, opened 2025-05-09
Fix upstream version in CSV description

lunarwhite:
As title, fix minor nits found while trying to build v1.16.0 operator locally.

### DIFF
--- a/config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
@@ -39,7 +39,7 @@ metadata:
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/os.linux: supported
-  name: cert-manager-operator.v1.15.0
+  name: cert-manager-operator.v1.16.0
   namespace: cert-manager-operator
 spec:
   apiservicedefinitions: {}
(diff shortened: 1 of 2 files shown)


config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -39,7 +39,7 @@ metadata:
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/os.linux: supported
-  name: cert-manager-operator.v1.15.0
+  name: cert-manager-operator.v1.16.0
```
swghosh, line 42: Can you change this to

```suggestion
  name: cert-manager-operator
```

That way we don't need to update this field in the CSV of `config/*/` for future bumps.

swghosh (approved):

swghosh:
/label docs-approved
/label px-approved
A follow-up over #268 

/retitle Fix upstream version in CSV description

config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
```diff
@@ -69,7 +69,7 @@ spec:
       name: certmanagers.operator.openshift.io
       version: v1alpha1
   description: |
-    The cert-manager Operator for Red Hat OpenShift provides seamless support for [cert-manager v1.15.5](https://github.com/cert-manager/cert-manager/tree/v1.15.2), which automates certificate management.
+    The cert-manager Operator for Red Hat OpenShift provides seamless support for [cert-manager v1.16.4](https://github.com/cert-manager/cert-manager/tree/v1.16.4), which automates certificate management.
```
lunarwhite, line 72: @swghosh What about this field? Do we still want to pin a specific version, or just pointing to https://github.com/cert-manager/cert-manager/releases?

swghosh: We want users to know from the Catalog page at the time of installation about which upstream version of cert-manager they're installing. There were such requests in the past which required us to make this doc effort of putting it in the CSV description: so unfortunately we cannot get rid of this yet!

lunarwhite: Makes sense

config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml
lunarwhite: I'm afraid we still need to keep `.v1.16.0` for consistency, since it's also be coded in L109 https://github.com/openshift/cert-manager-operator/blob/c1fb4fcbf97e42f1a0048735601f2f6136573d2b/config/manifests/bases/cert-manager-operator.clusterserviceversion.yaml#L109

PR #292, opened 2025-06-25
CM-538: CVE-2025-30204 cert-manager/jetstack-cert-manager-rhel9: jwt-go allows excessive memory allocation during header parsing [cert-manager-1.15]

PillaiManish:


go.mod
```diff
@@ -293,3 +293,8 @@ require (
 )
```
bharath-b-rh, line 293: Do we need this change? operator doesn't consume the package. And the pkg containing the vulnerabilities `github.com/cert-manager/cert-manager` is used in operator mainly for API defs and doesn't invoke any functionality involving jwt package, correct?

PR #294, opened 2025-07-01
Read Only Root Filesystem

jacobsee:
Add `readOnlyRootFilesystem: true` to embedded deployment yaml


jacobsee:
Wrong security context and already applied 🤦

PR #301, opened 2025-07-31
CM-607: Add e2es for customizing acme solver pods resources via controller flags

lunarwhite:


test/e2e/utils_test.go
```diff
@@ -764,3 +764,31 @@ func pollTillDeploymentAvailable(ctx context.Context, clientSet *kubernetes.Clie
 
 	return err
 }
+
+// VerifyContainerResources verifies that all containers in a pod have resources matching the expected configuration
+func VerifyContainerResources(pod corev1.Pod, expectedResources corev1.ResourceRequirements) error {
+	if len(pod.Spec.Containers) == 0 {
```
bharath-b-rh, line 770: nit: a pod in valid state will have atleast one container.

test/e2e/utils_test.go
```diff
@@ -764,3 +764,31 @@ func pollTillDeploymentAvailable(ctx context.Context, clientSet *kubernetes.Clie
...
+		return fmt.Errorf("no containers found in pod %s", pod.Name)
+	}
+
+	for _, container := range pod.Spec.Containers {
```
bharath-b-rh, line 774: nit: we don't have multiple containers as of today, maybe not in future too. But would it better to verify these on specific container using container name?

bharath-b-rh (approved):
/lgtm
/approve
/label docs-approved
/label px-approved

lunarwhite:
self adding
/label qe-approved

PR #327, opened 2025-10-16
Update OCP documentation links to latest versions

sebrandon1:
This PR updates OpenShift Container Platform documentation URLs to point to the latest available versions.

## Changes
- Automatically updated OCP documentation links using [ocp-doc-checker](https://github.com/sebrandon1/ocp-doc-checker)
- All documentation URLs now point to current versions

## Tool Information
This PR was created automatically by scanning the repository with `ocp-doc-checker`, which identifies and updates outdated OpenShift documentation links.

Repository: https://github.com/sebrandon1/ocp-doc-checker

---

**Tracking Issue:** https://github.com/sebrandon1/ocp-doc-checker/issues/18

### DIFF
--- a/docs/operand_metrics.md
+++ b/docs/operand_metrics.md
@@ -6,7 +6,7 @@ You can configure OpenShift Monitoring to collect metrics from cert-manager oper
 
 ### Enable User Workload Monitoring
 
-Cluster administrators can enable monitoring for user-defined projects by setting the `enableUserWorkload: true` field in the cluster monitoring ConfigMap object. For more details, Please look at the detailed documentation to [Configuring user workload monitoring](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/monitoring/configuring-user-workload-monitoring).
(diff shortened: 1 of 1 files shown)


bharath-b-rh:
> @bharath-b-rh Would you rather them be `latest` instead of pointing to a certain release? I can change it if so.

I feel so, since cert-manager releases aren't directly mapped to OCP releases. The latest always gets translated to the current release, and we wouldn't need to update it after every OCP release.

sebrandon1:
Updated with latest OCP documentation links (ran ocp-doc-checker again).

All documentation URLs now point to the latest available versions.

lunarwhite (changes requested):
@sebrandon1 first off thanks for your PR!

I assume we don't really need to worry about a version is not "latest" in this markdown doc. The equivalent content for customers reference is already there at https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/security_and_compliance/cert-manager-operator-for-red-hat-openshift#cert-manager-monitoring

My major concern is sometimes the RH docs paths would change between major releases. That means `https://docs.redhat.com/.../<a future version>/html/monitoring/configuring-user-workload-monitoring` might not always work. But what we could confirm is that `4.18` is not broken and won't be broken in the future.

I'd lean towards we keep it as-is to avoid bumping it every release and verifying if it's still valid. WDYT?

bharath-b-rh:
/lgtm
/approve
no user facing changes hence adding required labels
/label qe-approved
/label docs-approved
/label px-approved

PR #394, opened 2026-04-06
CM-873: Add comprehensive scenario-based e2e tests for the trust-manager

chiragkyal:
## Summary

Adds comprehensive scenario-based e2e tests for the trust-manager Bundle CR, covering the full lifecycle from Bundle creation through target sync verification under various TrustManager configurations.

### Test Coverage

**Group 1 — Default TrustManager (no optional features):**
- Inline / ConfigMap / Secret source → ConfigMap target
- Multiple sources, custom metadata, namespace selector filtering
- Target data drift reconciliation, source update propagation, Bundle deletion cleanup
- Negative: Secret target without SecretTargets enabled, useDefaultCAs without DefaultCAPackage enabled, ConfigMap source outside trust namespace

**Group 2 — SecretTargets enabled:**
- Inline / ConfigMap source → Secret target, dual ConfigMap + Secret targets
- Negative: Bundle name not in authorizedSecrets list

**Group 3 — DefaultCAPackage enabled:**
- useDefaultCAs → ConfigMap target, useDefaultCAs + Inline combined data

**Group 4 — SecretTargets + DefaultCAPackage enabled:**
- useDefaultCAs + Inline → ConfigMap + Secret dual targets

**Group 5 — Custom TrustNamespace:**
- ConfigMap source in custom trust namespace → ConfigMap target
- Negative: ConfigMap source in default namespace not synced when custom trust namespace is configured

**Group 6 — FilterExpiredCertificates enabled:**
- ConfigMap source with valid + expired certs → only valid cert in ConfigMap target
- Transition to Disabled → same Bundle re-syncs with both certs in target

PR #413, opened 2026-04-23
CM-873: e2e: use size comparison for CA bundle propagation assertion

chiragkyal:
## Summary

- Replaces the `ShouldNot(Equal(...))` assertion with a `BeNumerically(">", ...)` size comparison when verifying CNO has propagated the updated CA bundle to the operator namespace.
- Since the test appends a new certificate to the existing bundle, comparing the length is a more precise assertion that reflects the expected behavior (the bundle should grow), rather than just checking it changed.

## Closes
- https://github.com/openshift/cert-manager-operator/pull/394#discussion_r3128610500
- https://github.com/openshift/cert-manager-operator/issues/411
- https://github.com/openshift/cert-manager-operator/issues/410

/cc @bharath-b-rh 

## Summary by CodeRabbit

* **Tests**
  * Updated CA bundle propagation test verification logic to check for bundle size growth after configuration changes.


bharath-b-rh:
no user facing changes, hence adding all required labels
/label docs-approved
/label px-approved
/label qe-approved

PR #409, opened 2026-04-17
CM-966: Apply cluster TLS security profile to cert-manager operands

arun717:
# [CM-966](https://redhat.atlassian.net/browse/CM-966): Implement centralized TLS profile fetching and application

## Summary

Implements centralized TLS configuration for cert-manager operands by fetching and applying TLS settings from the cluster APIServer resource (`apiserver.config.openshift.io/cluster`). This ensures cert-manager components honor the cluster-wide TLS security profile.

Fixes: [CM-966](https://redhat.atlassian.net/browse/CM-966)
Related: [CM-954](https://redhat.atlassian.net/browse/CM-954)

### Core Implementation

1. **TLS Profile Package** (`pkg/tlsprofile/`)
   - Maps OpenShift TLS security profiles to cert-manager CLI flags
   - Supports Old, Intermediate (default), Modern, and Custom profiles
   - Converts OpenSSL cipher names to IANA format using library-go
   - Includes comprehensive unit tests

2. **TLS Profile Hook** (`pkg/controller/deployment/tls_profile_hook.go`)
   - Fetches TLS configuration from cluster APIServer resource
   - Applies appropriate flags based on deployment type:
     - `cert-manager-webhook`: main TLS + metrics TLS flags
     - `cert-manager` (controller): metrics TLS flags only
     - `cert-manager-cainjector`: metrics TLS flags only
   - Merges flags with override semantics (TLS settings take precedence)
   - Platform-agnostic: only active on OpenShift clusters

3. **Integration**
   - Registered hook in generic deployment controller
   - Added APIServer informer to watch for TLS profile changes
   - Ensures deployments reconcile when cluster TLS profile is updated

4. **RBAC**
   - Added permissions to read `apiservers` from `config.openshift.io` API group
   - Updated kubebuilder annotations, role.yaml, and CSV

### Testing

- Added comprehensive unit tests for all TLS profile types
- Tests cover: Intermediate, Modern, Old, Custom, and nil profiles
- Validates flag override behavior and multi-container deployment handling
- All tests passing ✅

### Known Limitations

- **Curve preferences**: Not configurable due to upstream cert-manager limitations. Operands inherit Go's default curve ordering for TLS 1.2/1.3 handshakes. Will be addressed when cert-manager adds explicit curve preference controls.
- **Trust-manager**: Currently does not support TLS configuration flags in upstream.

### Unit Tests
```bash
make test-unit
```

### Manual Testing (OpenShift Cluster)

1. Deploy cert-manager-operator with these changes
[...]

PR #407, opened 2026-04-14
CM-786: Updates controller override arg list to allow certificate-request-minimum-backoff-duration and fmt changes

bharath-b-rh:
The PR has following changes:
- A new feature is backported to downstream cert-manager to support certificate request backoff duration as part of https://github.com/openshift/jetstack-cert-manager/pull/36 and in this PR the arg is added to the supported list for users to override the default configuration.
- Includes `make fmt` suggested changes
- Includes `make update-vendor` suggeste changes

## Summary by CodeRabbit

* **New Features**
  * Added support for the --certificate-request-minimum-backoff-duration flag in cert-manager controller deployment overrides.

* **Tests**
  * Added test helpers to centralize informer/client setup and reduced duplicated test setup.
  * Expanded table-driven tests for deployment override validation, covering accepted/rejected args, scoping, error cases, and nil configs.

* **Style / Refactor**
  * Formatting and alignment improvements across test files.

PR #416, opened 2026-04-29
CM-966: Implement centralized TLS profile fetching and application without istio-csr changes

arun717:
Same description as [https://github.com/openshift/cert-manager-operator/pull/409]

Removed the changes dependent on istio-csr upstream changes.

## Summary by CodeRabbit

* **New Features**
  * Cert-manager operands now honor cluster TLS security profiles, applying TLS flags and cipher suites.
  * New client TLS configuration used for outbound HTTPS in tests and components.

* **Improvements**
  * Operator granted read access to cluster API server config to enable TLS policy enforcement.
  * Reconciliation now propagates request context for more reliable operations.

* **Documentation**
  * Added documented controller argument example to adjust certificate-request backoff duration.

* **Tests**
  * Comprehensive unit and e2e tests added/updated to validate TLS behavior and context propagation.


mytreya-rh:
@arun717 i wonder if we should take tlsAdherence into consideration after re-reading the EP.
https://github.com/richardsonnick/enhancements/blob/85a20f3cc787cd48e17cd5069b855c0d4aa4300e/enhancements/security/centralized-tls-config.md#tls-adherence-modes
@bharath-b-rh wdyt? would it be correct to use tlsAdherence and pass down centrally configured profile only in case of Strict mode?

api/operator/v1alpha1/certmanager_types.go
```diff
@@ -22,6 +22,7 @@ type CertManagerSpec struct {
 	//   - "--acme-http01-solver-nameservers="8.8.8.8:53,1.1.1.1:53"
 	//   - "--dns01-recursive-nameservers=8.8.8.8:53,1.1.1.1:53"
 	//   - "--dns01-recursive-nameservers-only"
+	//   - "--certificate-request-minimum-backoff-duration=30m"
 	//
```
mytreya-rh: This should have been part of https://github.com/openshift/cert-manager-operator/pull/407 right?
i would recommend a separate PR, even it trivial, as it would be easy to track changes :)

pkg/controller/istiocsr/controller.go
```diff
@@ -197,10 +197,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, fmt.Errorf("failed to update %q istiocsr.openshift.operator.io with finalizers: %w", req.NamespacedName, err)
 	}
 
-	return r.processReconcileRequest(istiocsr, req.NamespacedName)
+	return r.processReconcileRequest(ctx, istiocsr, req.NamespacedName)
 }
 
-func (r *Reconciler) processReconcileRequest(istiocsr *v1alpha1.IstioCSR, req types.NamespacedName) (ctrl.Result, error) {
+func (r *Reconciler) processReconcileRequest(ctx context.Context, istiocsr *v1alpha1.IstioCSR, req types.NamespacedName) (ctrl.Result, error) {
 	istioCSRCreateRecon := false
```
mytreya-rh: The ctx being passed down seems to be unused anywhere in the call chain?
Can we remove this (and other related) change(s) from the PR as anyways this is not touching IstioCSR?

pkg/controller/deployment/tls_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	certmanagerWebhookDeployment    = "cert-manager-webhook"
+	certmanagerCAinjectorDeployment = "cert-manager-cainjector"
+
+	argKeyValSeparator = "="
+)
+
+// mergeContainerArgs merges the source args with override values
+// using a map that tracks unique keys for each arg containing a
+// key value pair of form `key[=value]`.
+func mergeContainerArgs(sourceArgs []string, overrideArgs []string) (destArgs []string) {
```
mytreya-rh: This looks to be same as mergeContainerArgs from certmanager pkg deployment_helper.go.
Lets move this and other utility functions to a shared pkg (maybe common ?) and then be used from both deployment as well as certmanager pkg to avoid duplicate code and duplicate maintenance

pkg/controller/deployment/tls_profile_hook.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// WithClusterTLSProfileFromAPIServer applies TLS settings from
+// apiserver.config.openshift.io/cluster to cert-manager operands. It is only
+// registered when the shared config.openshift.io informer factory is available
+// (OpenShift clusters).
+func WithClusterTLSProfileFromAPIServer(apiServerInformer configinformersv1.APIServerInformer) func(*operatorv1.OperatorSpec, *appsv1.Deployment) error {
+	return func(_ *operatorv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		if len(deployment.Spec.Template.Spec.Containers) != 1 {
+			return nil
```
mytreya-rh: should an error be returned here? The return nil would mean a silent skip of everything else right?

pkg/controller/common/tls_profile_hook.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		effective, err := tlsprofile.EffectiveSpec(apiServer.Spec.TLSSecurityProfile)
+		if err != nil {
+			return err
+		}
+
+		adherence := apiServer.Spec.TLSAdherence
+		if !libgocrypto.ShouldHonorClusterTLSProfile(adherence) {
+			return nil
```
mytreya-rh, line 42: would it be good to log? or is it already logged in the called function?

mytreya-rh:
Hey @arun717 thanks for your patience :-) finally did a first pass, and left few comments.
In addition, lets structure the commits into:
1. Bumping the modules needed for the new API
2. Vendoring the new modules
3. Code Changes for the feature

That way it will be easier to do the subsequent review (vs 26 commits and 2k+ files ;)

api/operator/v1alpha1/certmanager_types.go
arun717: Done. Removed.

pkg/controller/istiocsr/controller.go
arun717: Removed the unused ctx from everywhere.

pkg/controller/deployment/tls_helpers.go
arun717: Done. Moved to common dir.

pkg/controller/deployment/tls_profile_hook.go
arun717: ok returning error in the modified version.

pkg/controller/common/tls_profile_hook.go
arun717: Adding V(4) log when tlsAdherence skips cluster TLS profile.
using this level logging since always-on logging would be too much logs(noise).

pkg/controller/certmanager/generic_deployment_controller.go
```diff
@@ -67,6 +68,9 @@ func newGenericDeploymentController(
 		))
 
 		informers = append(informers, infraInformerFactory.Config().V1().Infrastructures().Informer())
+
+		hooks = append(hooks, common.WithClusterTLSProfileFromAPIServer(infraInformerFactory.Config().V1().APIServers()))
+		informers = append(informers, infraInformerFactory.Config().V1().APIServers().Informer())
```
bharath-b-rh, line 72: Infrastructure CRD optional check was specifically added for MicroShift and it applies for APIServer as well currently.
But should we this with Infrastructure and instead have a dedicated check for APIServer CRD existence?
Maybe not required to handle now, we can wait for explicit request. Just stating the observation here.

pkg/controller/common/container_args.go
```diff
@@ -0,0 +1,69 @@
...
+func MergeContainerArgs(sourceArgs []string, overrideArgs []string) []string {
+	destArgMap := make(map[string]string)
+	ParseArgMap(destArgMap, sourceArgs)
+	ParseArgMap(destArgMap, overrideArgs)
+
+	destArgs := make([]string, len(destArgMap))
+	i := 0
+	for key, val := range destArgMap {
+		if len(val) > 0 {
+			destArgs[i] = fmt.Sprintf("%s%s%s", key, argKeyValSeparator, val)
```
bharath-b-rh: nit: Since it's a simple string derivation here, should we really use fmt here?
```suggestion
			destArgs[i] = key + argKeyValSeparator + val
```

pkg/controller/istiocsr/deployments_test.go
```diff
@@ -25,11 +25,12 @@ import (
...
-		preReq         func(*Reconciler, *fakes.FakeCtrlClient)
-		updateIstioCSR func(*v1alpha1.IstioCSR)
-		skipEnvVar     bool
-		wantErr        string
+		name             string
+		preReq           func(*Reconciler, *fakes.FakeCtrlClient)
+		updateIstioCSR   func(*v1alpha1.IstioCSR)
+		skipEnvVar       bool
+		wantErr          string
+		wantErrSubstring bool // if true, err must contain wantErr (for k8s validation text that differs by version)
```
bharath-b-rh: nit:
```suggestion
		wantErrSubstring bool
```

pkg/tlsprofile/client_config.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, err
+	}
+	curves := append([]tls.CurveID(nil), DefaultCurvePreferences...)
+	return &tls.Config{
+		RootCAs:            rootCAs,
+		MinVersion:         minVer,
+		CipherSuites:       cipherIDs,
+		CurvePreferences:   curves,
+		InsecureSkipVerify: false,
+	}, nil
```
bharath-b-rh: nit: Since go ignores CipherSuites for TLS 1.3, and instead uses it's preferred list, should we short-circuit here when min version is 1.3?

```suggestion
	tlsConfig := &tls.Config{
		RootCAs:            rootCAs,
		MinVersion:         minVer,
		CurvePreferences:   append([]tls.CurveID(nil), DefaultCurvePreferences...),
		InsecureSkipVerify: false,
	}
	if minVer == tls.VersionTLS13 {
		return tlsConfig, nil
	}
	iana := libgocrypto.OpenSSLToIANACipherSuites(spec.Ciphers)
	if len(spec.Ciphers) > 0 && len(iana) == 0 {
		return nil, fmt.Errorf("no cipher suites after OpenSSL->IANA mapping")
	}
	cipherIDs, err := cipherSuiteIDsFromIANANames(iana)
	if err != nil {
		return nil, err
	}
	tlsConfig.CipherSuites = cipherIDs
	return tlsConfig, nil
```

pkg/controller/common/tls_profile_hook.go
```diff
@@ -0,0 +1,72 @@
...
+	certmanagerCAinjectorDeployment = "cert-manager-cainjector"
+)
+
+// WithClusterTLSProfileFromAPIServer applies TLS settings from
+// apiserver.config.openshift.io/cluster to cert-manager operands. It is only
+// registered when the shared config.openshift.io informer factory is available
+// (OpenShift clusters).
+func WithClusterTLSProfileFromAPIServer(apiServerInformer configinformersv1.APIServerInformer) func(*operatorv1.OperatorSpec, *appsv1.Deployment) error {
+	return func(_ *operatorv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		if len(deployment.Spec.Template.Spec.Containers) != 1 {
```
bharath-b-rh, line 29: Should we be enforcing this? One scenario I could think of where this would break is ServiceMesh configured clusters, where it could inject a sidecar and this would fail. Should we instead rely on the container names?

```
for _, name := range deployment.Spec.Template.Spec.Containers {
  switch name {
  case "cert-manager-controller", "cert-manager-cainjector", "cert-manager-webhook":
     MergeContainerArgs()
  }
}
```

pkg/controller/certmanager/generic_deployment_controller.go
```diff
@@ -67,6 +68,9 @@ func newGenericDeploymentController(
 		))
 
 		informers = append(informers, infraInformerFactory.Config().V1().Infrastructures().Informer())
+
+		hooks = append(hooks, common.WithClusterTLSProfileFromAPIServer(infraInformerFactory.Config().V1().APIServers()))
```
bharath-b-rh, line 71: Should `WithClusterTLSProfileFromAPIServer` be called before calling `withContainerArgsOverrideHook` and `withUnsupportedArgsOverrideHook`.
Currently,  `withContainerArgsOverrideHook` doesn't allow these ARGs to be overrided, but if its opened up in future...
But yes with `withUnsupportedArgsOverrideHook`, I think we should support overriding it, unless TLS compliance states explicitly states operators shouldn't allow overriding OpenShift cluster configuration.

pkg/tlsprofile/client_config_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"crypto/x509"
+	"strings"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func TestClientTLSConfig_intermediate(t *testing.T) {
```
bharath-b-rh: I think all tests in this file should be one test suite in go test table format for better readability. Also we could add a test for Modern profile type.

test/e2e/utils_test.go
```diff
@@ -1467,21 +1475,29 @@ func isPodReady(pod *corev1.Pod) bool {
...
 }
 
-// httpsGetCallWithCA performs an HTTPS GET request with custom CA certificate
-func httpsGetCallWithCA(url string, caCertPEM []byte) error {
-	// Create a certificate pool and add the CA cert
+// httpsGetCallWithCA performs an HTTPS GET request with custom CA certificate.
+// The client's minimum TLS version, cipher suites, and curve preferences match
+// apiserver.config.openshift.io/cluster spec.tlsSecurityProfile (via
+// pkg/tlsprofile), consistent with cluster-wide TLS policy.
+func httpsGetCallWithCA(ctx context.Context, url string, caCertPEM []byte) error {
```
bharath-b-rh, line 1483: We should add a test case to validate the containers indeed have TLS args set and matches what's configured on the cluster.

pkg/controller/certmanager/generic_deployment_controller.go
arun717: For now not making any code changes for this.

pkg/controller/common/container_args.go
arun717: done.

pkg/controller/istiocsr/deployments_test.go
arun717: done.

pkg/tlsprofile/client_config.go
arun717: done.

pkg/controller/common/tls_profile_hook.go
arun717: Thanks for raising this. It is a fair thing to think about for mesh-enabled clusters.  This does not look like a Service Mesh issue in practice. 
Deployment hooks run on the embedded manifest before apply, not on the live cluster object. 
Mesh sidecars are injected at admission after apply, so reconcile still sees one container here.

The hook always patches Containers[0], so the len != 1 check guards that assumption. It fails fast if the manifest is empty or has multiple containers, rather than patching the wrong one or panicking. 

That matches how the other cert-manager deployment hooks work (log level, proxy, overrides, credentials), and our bindata Deployments always ship a single operand container per Deployment.
Please do let me know if I have missed something.

pkg/controller/certmanager/generic_deployment_controller.go
arun717: Good question on hook ordering.

We’re keeping TLS last intentionally. Moving it before the override hooks would let UnsupportedConfigOverrides override cluster TLS, which conflicts with StrictAllComponents (“all components must honor the cluster TLS profile”).

On the two override paths:
- withContainerArgsOverrideHook: supported-args validation already blocks TLS flags today. Even if those flags were allowlisted later, TLS-last would still ensure the cluster profile wins.
- withUnsupportedArgsOverrideHook: could set TLS flags today, but with TLS running last, cluster TLS settings take precedence which is consistent with strict adherence.

This aligns with the downstream-operator pattern of applying apiserver/cluster TLS via operand flags. Happy to revisit only if product explicitly decides users may override cluster TLS under strict adherence.

pkg/tlsprofile/client_config_test.go
arun717: done.

test/e2e/utils_test.go
arun717: done.

pkg/controller/certmanager/generic_deployment_controller.go
bharath-b-rh: The [Centralized TLS Config](https://github.com/openshift/enhancements/blob/master/enhancements/security/centralized-tls-config.md) EP explicitly states below precedence, and it should be considered. In some scenarios if the configured TLS policy is causing issues for a  cert-manager provider, instead of asking to modify the central config, we should provide way to mitigate it locally. And initially I was thinking to suggest using `withUnsupportedArgsOverrideHook`, but I think we should support it in `withContainerArgsOverrideHook`.

```
Override Precedence:
- Component-specific CR configuration (e.g., IngressController.spec.tlsSecurityProfile) takes highest precedence when explicitly set
- If no component-specific override is set, the cluster-wide apiserver.config.openshift.io/cluster configuration applies
```

pkg/controller/common/tls_profile_hook.go
bharath-b-rh: Yeah, this was miss from my end, didn't apply the static manifests logic we have.
But I still think it would be good to check against container names instead of hardcoding to keep it generic, as this would add to the maintenance if container count changes in future. But that's fine, it can be taken up when any such usecase comes up

arun717: Sure, not making any change for now.

pkg/controller/certmanager/generic_deployment_controller.go
arun717: The response on this question from forum:
```
If a component has its own APIs that configure TLS server settings, and those APIs are set, the component should honor its own config APIs over the cluster-wide configuration.
But if those configurations are unset or reference "Default" settings, then the cluster-wide settings should be the next set of configuration that is used.
```
So making the necessary changes and keeping withUnsupportedArgsOverrideHook last.

test/e2e/tls_profile_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Cluster TLS security profile", Label("Platform:Generic"), Ordered, func() {
```
bharath-b-rh: Don't we need `TechPreview` label, since the feature depends on `TLSAdherence` featuregate?


```suggestion
var _ = Describe("Cluster TLS security profile", Label("Platform:Generic", "TechPreview"), Ordered, func() {
```

test/e2e/tls_profile_test.go
```diff
@@ -0,0 +1,50 @@
...
+	})
+
+	It("should configure operand container TLS args from apiserver cluster profile", func() {
+		state, err := getClusterTLSProfileState(ctx)
+		if apierrors.IsNotFound(err) {
+			Skip("apiserver.config.openshift.io/cluster is not available on this cluster")
+		}
+		Expect(err).NotTo(HaveOccurred(), "failed to read cluster TLS profile state")
+
+		if !state.honor {
```
bharath-b-rh: By default the TLSAdherence and TLSProfile won't be set on a cluster. Shouldn't we have a testcase to patch it and also check if the TLS args are set? 

IIUC, the testcase here checks the legacy adherence scenario alone, correct?

test/e2e/tls_profile_test.go
arun717: Good catch. Applying the suggestion.

test/e2e/tls_profile_test.go
arun717: On a fresh TechPreview cluster tlsAdherence and tlsSecurityProfile are usually unset, so the old test would almost always skip and never validate the operator behavior.

The test now:
Saves the current apiserver.config.openshift.io/cluster TLS settings
Patches the cluster to tlsAdherence: StrictAllComponents and tlsSecurityProfile.type: Modern (TLS 1.3 — a clear, distinct profile)
Waits and verifies operand deployments get the expected TLS args via verifyOperandTLSArgsMatchClusterProfile
Restores the original settings in DeferCleanup (if tlsAdherence was unset, it restores to LegacyAdheringComponentsOnly since the field cannot be removed once set)

bharath-b-rh:
/lgtm
/approved

Please check the [comment](https://redhat.atlassian.net/browse/CM-1102?focusedCommentId=17254012) for pre-merge validations.
/label qe-approved

/label px-approved

cc: @snarayan-redhat for docs-approved. Please help with the release notes.

```
cert-manager Operator applies apiserver.config.openshift.io/cluster TLS settings to cert-manager, webhook, and cainjector deployments when tlsAdherence: StrictAllComponents is set (requires TLSAdherence feature gate). tlsSecurityProfile alone does not affect operands. unsupportedConfigOverrides is applied after the cluster TLS hook so administrators can override TLS flags in unsupported break-glass scenarios.

```

/hold for @mytreya-rh review.

pkg/tlsprofile/tlsprofile.go
mytreya-rh, line 1: the tlsprofile seems to be used only from pkg/controller/common.
Can this folder not or it contents not be moved to the same location?
(maybe we can take it up later as well)

PR #427, opened 2026-05-29
CM-1043: Add IstioCSR e2e coverage and IstioCSR-ServiceMesh smoke tests

arun717:
Adds end-to-end coverage for IstioCSR reconciler and operand behavior that was not covered by existing istio_csr_test.go specs, plus two opt-in OpenShift Service Mesh (OSM) smoke tests.

New file test/e2e/istio_csr_p0_test.go introduces 12 P0 scenarios under Feature:IstioCSR / Platform:Generic, covering:
- CR validation (singleton name, multi-instance rejection)
- Issuer paths (ClusterIssuer happy path, unsupported ACME negative)
- Operand configuration (custom gRPC port, log level/format + deployment drift reconciliation)
- Managed resource reconciliation (ServiceAccount, Service, NetworkPolicy)
- Status contract (IstioCSRImage, gRPC endpoint, SA, ClusterRole/Binding)
- User-provided CA ConfigMap negatives (missing key, missing namespace)

Also adds an OSM smoke context (Feature:ServiceMesh) with two cases that auto-skip when no istiod control plane is present:

- Maistra-labeled member namespace receives istio-ca-root-cert
- gRPC CreateCertificate succeeds for a mesh workload SPIFFE identity

Supporting changes:

utils_test.go: waitForIstioCSRConditionMessage helper for degraded-condition assertions
istio_csr_test.go: traceability label on existing grpc smoke (CM-867-TC-010)

Test plan

Default e2e (no mesh required):
```E2E_GINKGO_LABEL_FILTER='Feature:IstioCSR && Platform:Generic' make test-e2e```

Run P0 specs individually if needed:
```E2E_GINKGO_LABEL_FILTER='ISTIOCSR-P0-001' make test-e2e```

OSM smoke (cluster with ready istiod required):
```E2E_GINKGO_LABEL_FILTER='Feature:IstioCSR && Feature:ServiceMesh' make test-e2e```

Confirm existing IstioCSR grpc spec still passes:
```E2E_GINKGO_LABEL_FILTER='CM-867-TC-010' make test-e2e```

Verify OSM smokes are skipped cleanly on clusters without Service Mesh / istiod

Notes

- OSM tests use maistra.io/member-of=<istiod-namespace> and discover the control-plane namespace dynamically.
- IstioCSR CR is created in a dedicated test namespace while spec.istioCSRConfig.istio.namespace targets the discovered istiod namespace.
- No production code changes; test-only diff under test/e2e/.

## Summary by CodeRabbit

* **Tests**
  * Added extensive e2e coverage for Istio-CSR: validation, reconciliation across issuers, degraded states, resource recovery/drift correction, status contract checks, OpenShift Service Mesh smoke tests, and ACME HTTP-01 ingress handling. Improved stability with targeted timeouts, polling, and clearer auth error reporting.

* **Chores**
  * Pinned Vault test image, hardened Vault setup/command handling, and tightened test invocation quoting for CI reliability.

* **Documentation**
  * Added a detailed e2e test plan with 10 scenarios, coverage mapping, and run guidance.

### CODE REVIEW

File: test/e2e/plans/pr-379/test-cases.md [hunk 1]
```diff
@@ -0,0 +1,255 @@
+# Test Plan: CM-867 — TrustManager operand reconcilers (PR #379)
```
bharath-b-rh: I don't think we need to include this file.

test/e2e/testdata/istio/grpcurl_job_with_cluster_id.yaml
```diff
@@ -7,6 +7,8 @@ spec:
   completions: 1
   template:
     metadata:
+      annotations:
```
bharath-b-rh: Apparently this annotation has been deprecated and instead is now a label.

Ref: [https://istio.io/latest/docs/reference/config/annotations](https://istio.io/latest/docs/reference/config/annotations/#:~:text=Specifies%20whether%20or%20not%20an%20Envoy%20sidecar%20should%20be%20automatically%20injected%20into%20the%20workload.%20This%20annotation%20has%20been%20deprecated%20in%20favor%20of%20the%20sidecar.istio.io/inject%20label%20documented%20here. )

test/e2e/testdata/servicemesh/istio-csr-operand.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: operator.openshift.io/v1alpha1
```
bharath-b-rh: There is `test/e2e/testdata/istio/istio_csr_template.yaml` and this too. Both have different config, but for same purpose. I think it's good to have just one and templatise fields to keep it generic.

test/e2e/condition_matcher_test.go
```diff
@@ -102,6 +102,9 @@ func verifyOperatorStatusCondition(client v1alpha1client.OperatorV1alpha1Interfa
 			if apierrors.IsNotFound(err) {
 				return false, nil
 			}
+			if apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err) {
+				return false, fmt.Errorf("cannot get certmanagers.operator.openshift.io/cluster (run 'oc login' and 'oc get certmanager cluster'): %w", err)
```
bharath-b-rh: The error message and the error doesn't match. Also is it really required to handle these errors?

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	istioCSRP0ClusterIssuerName          = "selfsigned-issuer"
```
bharath-b-rh: nit: using `P0` in names looks odd. Could we update in all?

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+	istioCSRP0CAClusterIssuerName        = "istiocsr-p0-ca-clusterissuer"
+	istioCSRP0CASecretName               = "root-secret"
+	istioCSRP0CACertificateName          = "my-selfsigned-ca"
+	istioCSRP0RejectAnnotation           = "operator.openshift.io/istio-csr-reject-multiple-instance"
+	istioCSRP0ManagedResourceLabel       = "app.kubernetes.io/name=cert-manager-istio-csr"
+	istioCSRP0ManagedByLabel             = "app.kubernetes.io/managed-by"
+	istioCSRP0ManagedByExpectedValue     = "cert-manager-operator"
+	istioCSRP0ManagedAppLabel            = "app"
+	istioCSRP0ManagedAppExpectedValue    = "cert-manager-istio-csr"
```
bharath-b-rh: These don't have usage, are these required?

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+	addServerBlock                bool
+	addIstioDataPlaneSelector     bool
+	addCustomCAConfigMap          bool
+	addControllerConfigLabels     bool
+}
+
+func generateMeshWorkloadCSR(meshNamespace, serviceAccountName string) string {
+	csrTemplate := &x509.CertificateRequest{
+		Subject: pkix.Name{
+			Organization: []string{"OpenShift Service Mesh E2E"},
```
bharath-b-rh: nit :) 
```suggestion
			Organization: []string{"OpenShift cert-manager IstioCSR E2E"},
```

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+	copied := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: targetNS,
+		},
+		Data: source.Data,
+		Type: source.Type,
+	}
+	_, err = clientset.CoreV1().Secrets(targetNS).Create(ctx, copied, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
```
bharath-b-rh: `IsAlreadyExists` is not considered error here. Should `Update` be used instead of `Create`. Or is the secret cleaned up in `After` block should already exist be an error. This is used for copy the TLS secret, so I believe we shouldn't assume secret already existing as the required data, which could cause failures and probably difficult to debug in CI env.

Already exist handling should be revisited in other places as well.

test/e2e/istio_csr_operand_test.go
```diff
@@ -0,0 +1,837 @@
...
+		loader.CreateFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_ca_issuer.yaml"), namespace)
+
+		DeferCleanup(func() {
+			loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "istio", "istio_ca_issuer.yaml"), namespace)
+			loader.DeleteFromFile(testassets.ReadFile, filepath.Join("testdata", "self_signed", "certificate.yaml"), namespace)
+		})
+	}
+
+	newIstioCSR := func(namespace string, cfg istioCSRP0Config) *unstructured.Unstructured {
+		issuerRefKind := "Issuer"
```
bharath-b-rh, line 140: Nice to have the generic code, but I don't think we are creating the resources accordingly too correct?

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func() {
+			err := loader.DynamicClient.Resource(istiocsrSchema).Namespace(namespace).Delete(ctx, istioCSR.GetName(), metav1.DeleteOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				Expect(err).NotTo(HaveOccurred())
+			}
+		})
+	}
+
+	waitForIstioCSRReady := func(namespace string) {
```
bharath-b-rh: Would it helpful to create helpers function instead of first-class functions for reusability.

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+			customCAConfigMapName:      "external-ca",
+			customCAConfigMapNamespace: "non-existent-istiocsr-ca-ns",
+			customCAConfigMapKey:       "ca.crt",
+		})
+		createIstioCSR(ns.Name, istioCSR)
+		By("waiting for IstioCSR Degraded=True with missing CA ConfigMap namespace message")
+		Expect(waitForIstioCSRConditionMessage(ctx, loader, ns.Name, istioCSRP0ISTIOCSRName, v1alpha1.Degraded, metav1.ConditionTrue, "failed to fetch CA certificate ConfigMap", highTimeout, slowPollInterval)).NotTo(HaveOccurred())
+	})
+
+	Context("OpenShift Service Mesh smoke", Label("Feature:ServiceMesh"), Ordered, func() {
```
bharath-b-rh: nit:

```suggestion
	Context("OpenShift Service Mesh smoke", Label("Feature:IstioCSR-ServiceMesh"), Ordered, func() {
```

test/e2e/servicemesh_helpers_test.go
```diff
@@ -0,0 +1,655 @@
...
+		Version:  "v1",
+		Resource: "istios",
+	}
+)
+
+// deriveClusterID returns the Istio multi-cluster ID derived from the API server URL.
+// Example: https://api.bhb.gcp.devcluster.openshift.com:6443 -> api-bhb-gcp-devcluster-openshift-com:6443
+func deriveClusterID(cfg *rest.Config) string {
+	host := strings.TrimPrefix(cfg.Host, "https://")
+	host = strings.TrimPrefix(host, "http://")
```
bharath-b-rh:
```suggestion
	host = strings.TrimPrefix(cfg.Host, "http://")
```

test/e2e/servicemesh_helpers_test.go
```diff
@@ -0,0 +1,655 @@
...
+	_, err := loader.DynamicClient.Resource(clusterExtensionGVR).List(ctx, metav1.ListOptions{Limit: 1})
+	return err == nil
+}
+
+func ossmInstallEnabled() bool {
+	return os.Getenv("E2E_INSTALL_SERVICE_MESH") != "false"
+}
+
+func ossmIstioVersion() string {
+	if v := os.Getenv("E2E_OSM_ISTIO_VERSION"); v != "" {
```
bharath-b-rh, line 150: Should we define these versions in Makefile, so that during rebases these too gets updated.

test/e2e/servicemesh_helpers_test.go
```diff
@@ -0,0 +1,655 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ossmIstioSystemCACertName,
+			Namespace: ossmIstioSystemNamespace,
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			CommonName: ossmIstioSystemCACertName,
+			SecretName: ossmIstioSystemCASecretName,
+			IsCA:       true,
+			Duration:   &metav1.Duration{Duration: 2 * time.Hour},
+			IssuerRef: certmanagermetav1.ObjectReference{
```
bharath-b-rh: ObjectReference has been deprecated.
```suggestion
			IssuerRef: certmanagermetav1.IssuerReference{
```

test/e2e/issuer_vault_test.go
```diff
@@ -117,9 +117,12 @@ var _ = Describe("Vault Issuer", Ordered, Label("Platform:Generic"), func() {
 		})
 
 		By("setting up Vault server in cluster with PKI secrets engine enabled")
+		setupCtx, setupCancel := context.WithTimeout(context.Background(), vaultSetupTimeout)
```
bharath-b-rh, line 118: The vaultSetupTimeout is higher than the highTimeout. highTimeout is set on the suite, this will cause is it not?
```
	highTimeout      = 10 * time.Minute
	vaultSetupTimeout = 30 * time.Minute
	vaultPodStartTimeout = 10 * time.Minute
```

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+
+		istioCSR := newIstioCSR(ns.Name, istioCSRP0Config{
+			addServerBlock: true,
+			serverPort:     7443,
+		})
+		createIstioCSR(ns.Name, istioCSR)
+		waitForIstioCSRReady(ns.Name)
+
+		status := getIstioCSRStatus(ns.Name)
+		endpoint := status["istioCSRGRPCEndpoint"].(string)
```
bharath-b-rh: Should the assertion be validated here? And use `Eventually` to poll the value is populated.

```suggestion
		if endpoint, ok := status["istioCSRGRPCEndpoint"].(string); !ok {
		}
```

test/e2e/istio_csr_operand_test.go
```diff
@@ -0,0 +1,837 @@
...
+					succeededPodName = pod.Name
+				}
+			}
+			Expect(succeededPodName).NotTo(BeEmpty())
+
+			logStream, err := clientset.CoreV1().Pods(meshMemberNS.Name).GetLogs(succeededPodName, &corev1.PodLogOptions{}).Stream(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			defer logStream.Close()
+
+			logData, err := io.ReadAll(logStream)
```
bharath-b-rh, line 721: Should we instead use `json.NewDecoder` to scan line-by-line to parse the logs effectively?

test/e2e/istio_csr_p0_test.go
```diff
@@ -0,0 +1,837 @@
...
+		SignatureAlgorithm: x509.SHA256WithRSA,
+	}
+	csr, err := library.GenerateCSR(csrTemplate)
+	Expect(err).NotTo(HaveOccurred())
+	return csr
+}
+
+func copySecretToNamespace(ctx context.Context, clientset *kubernetes.Clientset, sourceNS, targetNS, secretName string) {
+	source, err := clientset.CoreV1().Secrets(sourceNS).Get(ctx, secretName, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
```
bharath-b-rh: Should we be using `Expect` outside gingko blocks?

test/e2e/istio_csr_operand_test.go
```diff
@@ -0,0 +1,837 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
```
bharath-b-rh, line 31: Should we move all the constants to a dedicated file?

test/e2e/servicemesh_helpers_test.go
```diff
@@ -0,0 +1,655 @@
...
+	_, err := clientset.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+	}, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
+}
+
+func applyManifestsFromFile(ctx context.Context, loader library.DynamicResourceLoader, clientset *kubernetes.Clientset, assetFn func(string) ([]byte, error), filename string) error {
```
bharath-b-rh, line 173: Should we be reusing `library.CreateFromFile()` instead

test/e2e/plans/pr-379/test-cases.md
arun717: Removed.

test/e2e/testdata/istio/grpcurl_job_with_cluster_id.yaml
arun717: Updated Istio grpcurl Job manifests to use the label instead of the deprecated annotation

test/e2e/testdata/servicemesh/istio-csr-operand.yaml
arun717: Made the changes. Unified template: testdata/istio/istio_csr_template.yaml

It now supports two profiles via Profile:
"minimal" (default) - previous istio_csr_template.yaml behavior
"ossm" - previous istio-csr-operand.yaml behavior

test/e2e/condition_matcher_test.go
arun717: Removing this part since E2E already assumes a working cluster client from BeforeSuite (config.GetConfig()). If Get fails with 401/403 here, the raw apiserver error is enough.

test/e2e/istio_csr_p0_test.go
arun717: Removed P0 from IstioCSR e2e identifiers across the package.

test/e2e/istio_csr_p0_test.go
arun717: They are likely leftovers from a planned label assertion that was never implemented. Removing them.

test/e2e/istio_csr_p0_test.go
arun717: done.

test/e2e/istio_csr_p0_test.go
arun717: Have added a shared upsertSecret helper in most places where IsAlreadyExists and secret-create patterns were found.

test/e2e/istio_csr_operand_test.go
arun717: Removed the unused generic branches from istioCSRBuildConfig and newIstioCSR
Removed fields:

- istioControlPlaneNamespace
- istioDataPlaneSelector / addIstioDataPlaneSelector
- controllerConfigLabels / addControllerConfigLabels

test/e2e/istio_csr_p0_test.go
arun717: Extracted shared IstioCSR ready-wait logic into istio_csr_helpers_test.go

test/e2e/istio_csr_p0_test.go
arun717: done.

test/e2e/servicemesh_helpers_test.go
arun717: No changes needed actually.

The two lines are meant to run in sequence on the current value, not both on the raw cfg.Host:

Strip https:// if present.
Strip http:// from whatever remains.

What breaks if line 78 uses cfg.Host is
host := strings.TrimPrefix(cfg.Host, "https://")
host = strings.TrimPrefix(cfg.Host, "http://")  // overwrites line 1

I will replace it with something like this for better readability
host := strings.TrimPrefix(strings.TrimPrefix(cfg.Host, "https://"), "http://")

test/e2e/servicemesh_helpers_test.go
arun717: sure. It will be consistent with how other versions are mentioned as of now.

test/e2e/servicemesh_helpers_test.go
arun717: done

test/e2e/issuer_vault_test.go
arun717: vaultSetupTimeout (30m) is intentionally longer than highTimeout (10m) for slow Vault setup. It doesn’t conflict with the 120m Ginkgo suite timeout or 2h go test timeout. 

However there is one gap that ctx is created at the start of BeforeEach with a 10m deadline, while setup runs on a separate 30m setupCtx.
So, if setup takes >10m, ctx is dead before the It blocks run. We should start or refresh ctx after setup completes. 
Making that change in the test.

test/e2e/istio_csr_p0_test.go
arun717: Eventually isn’t needed after expectIstioCSROperandReady. I’ll use the typed IstioCSRStatus return instead of map type asserts.

test/e2e/istio_csr_operand_test.go
arun717: grpcurl emits a single JSON response object, not NDJSON, so line-by-line scanning would fail on pretty-printed output. So ReadAll + Unmarshal seems appropriate here.

test/e2e/istio_csr_p0_test.go
arun717: What has been changed:
return error from copySecretToNamespace and Expect at the two call sites, to align with istio_csr_helpers_test.go and servicemesh_helpers_test.go.

test/e2e/istio_csr_operand_test.go
arun717: For this e2e package, what I observed was to:
Keep test-local constants close to the test file (like most in istio_csr_operand_test.go).

The current layout matches the rest of test/e2e. 
Let me know if you still want these constants to be moved.

test/e2e/servicemesh_helpers_test.go
arun717: I think its better to keep applyManifestsFromFile() in servicemesh_helpers_test.go for OSSM install flow.

CreateFromFile is a method on DynamicResourceLoader (loader.CreateFromFile(...)), and it’s great for simple fixture applies, but this path needs different behavior:

- installOSSMv3 wants error returns and wrapped context per step.
- CreateFromFile is test-fail oriented (require.NoError) and doesn’t return an error.
- applyManifestsFromFile has special handling for mixed objects in OSSM manifests (Namespace, ServiceAccount, ClusterRoleBinding, plus CRs).

test/e2e/istio_csr_operand_test.go
bharath-b-rh: The underlying for suggesting json decoder was scenarios where there extra content appended or incomplete json data due to halfway pod termination, where json unmarshaling could fail.

Probably the logic could be updated to something like

```
lines := bytes.Split(bytes.TrimSpace(logData), []byte("\n"))
var entry LogEntry
for i := len(lines) - 1; i >= 0; i-- {
    if json.Unmarshal(lines[i], &entry) == nil {
        break
    }
}
Expect(entry.CertChain).NotTo(BeEmpty())
```

arun717: done.

lunarwhite:
Suggestion: @arun717 would you please squash the commits before merging (by adding a comment `/label tide/merge-method-squash` or do it manually, I suggest the former one to avoid re-triggering next round of CIs). It's to maintain a clear git commits history for this repo. 17 back-and-forth edits without squash looks a little scary. Upstream guideline: https://github.com/kubernetes/community/blob/main/contributors/guide/github-workflow.md#squash-commits

bharath-b-rh:
The PR adds only e2e test cases, and no user facing changes 
Hence adding the label based on CI results. 
/label qe-approved

PR #429, opened 2026-06-04
CM-1042: To add multi-operand qualification suite in e2e tests

arun717:
## Summary

Adds an ordered Ginkgo e2e suite for [CM-786](https://redhat.atlassian.net/browse/CM-786) qualification when TrustManager and IstioCSR are enabled and installed at the same time. The suite patches the operator subscription to turn on both addon feature gates, creates both CRs concurrently, then runs baseline cert-manager sanity checks and automatable [CM-786](https://redhat.atlassian.net/browse/CM-786) doc steps (5–14).

Labels: Platform:Generic, Feature:MultipleOperands, TechPreview
https://docs.google.com/document/d/1bKI2OP7cZPVkGMWGhqEgJ6cCM35r2pUs7lqfDClUD9A/edit?tab=t.ovzsdk8nx64c#heading=h.edvlw82015i9

Run:
```make test-e2e E2E_GINKGO_LABEL_FILTER='Platform:Generic && Feature:MultipleOperands'```

Prerequisites: TechPreview cluster with cert-manager operator installed; OpenShift Service Mesh optional (steps 6–7 skip if absent).

Tests (Context: with operands installed concurrently)

| # | Spec | Verifies |
|---|------|----------|
| 1 | TrustManager and IstioCSR CRs ready after concurrent creation | TrustManager/cluster Ready + image; IstioCSR/default available with gRPC endpoint and image |
| 2 | CM-01: all operand deployments have desired replicas ready | Controller, webhook, cainjector, trust-manager, and namespace-scoped cert-manager-istio-csr deployments Available |
| 3 | CM-02: operand pods have no CrashLoopBackOff | No CrashLoopBackOff in operand + IstioCSR namespaces |
| 4 | CM-03: cert-manager webhook service has endpoints | Webhook Endpoints non-empty; webhook pods present |
| 5 | CM-10: issues a self-signed namespaced certificate | ClusterIssuer + Certificate → Ready; TLS secret with tls.crt / tls.key |
| 6 | CM-12: certificate with bogus issuer ref stays not Ready and does not break operands | Certificate Ready=False; target secret not created; operands and operator conditions still healthy |
| 7 | CM-786 step 5: IstioCSR operand resources exist with operator-managed labels | SA, Service, Deployment, ClusterRole with app.kubernetes.io/managed-by=part-of operator labels |
| 8 | CM-786 step 10: TrustManager qualification configuration is applied | Default CA ConfigMap; deployment args (--secret-targets-enabled, --filter-expired-certificates, --default-package-location); CR spec matches qualification policy |
| 9 | CM-786 step 10: Bundle propagates ConfigMap source to Secret in selected namespaces | Bundle syncs CM → Secret in namespace with istio-injection=enabled |
| 10 | CM-786 step 11: managed ClusterRoles are recreated after deletion | Trust-manager ClusterRole, IstioCSR ClusterRole, and cert-manager-controller-challenges (if present) recreated; operands stay Ready |
| 11 | CM-786 step 12: updating all three operator CRs keeps operands healthy | Label updates on CertManager, TrustManager, IstioCSR; operator conditions + deployments + operand readiness |
| 12 | CM-786 step 13: IstioCSR API updates reconcile successfully | logLevel 3 → 2 → 1 on IstioCSR CR; deployment --log-level args match each update |
| 13 | CM-786 step 14: IstioCSR gRPC CreateCertificate returns a cert chain | grpcurl Job succeeds; cert chain validates against test CA |
| 14 | CM-786 steps 6–7: Service Mesh uses istio-csr when Service Mesh is installed | If sailoperator.io/v1 Istio/default exists: spec.values.global.caAddress references istio-csr; istio-csr Service has endpoints — otherwise Skip |

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive end-to-end suite for multi-operand scenarios: operand readiness, pod/webhook health, certificate issuance and invalid-issuer behavior, bundle propagation, role recreation resilience, config updates without downtime, gRPC cert issuance, and optional Service Mesh checks.
  * Improved Vault e2e setup and diagnostics: dedicated setup/startup timeouts, richer pod failure reporting, and safer Vault command execution helpers.
* **Chores**
  * Pinned Vault test image tag, removed privileged flag from installer container spec, and updated test module dependencies.

### DIFF
--- a/test/e2e/issuer_acme_http01_test.go
+++ b/test/e2e/issuer_acme_http01_test.go
@@ -265,7 +265,7 @@ var _ = Describe("ACME Issuer HTTP01 solver", Label("Platform:Generic"), Ordered
 			}
 
 			err := wait.PollUntilContextTimeout(ctx, fastPollInterval, lowTimeout, true, func(ctx context.Context) (bool, error) {
-				pods, err := k8sClientSet.CoreV1().Pods("").List(ctx, metav1.ListOptions{
+				pods, err := k8sClientSet.CoreV1().Pods(ns.Name).List(ctx, metav1.ListOptions{
 					LabelSelector: acmeSolverPodLabel,
 				})
 				if err != nil {
@@ -286,10 +286,6 @@ var _ = Describe("ACME Issuer HTTP01 solver", Label("Platform:Generic"), Ordered
 				return false, nil // No matching pods yet, keep waiting
 			})
 			Expect(err).NotTo(HaveOccurred(), "should find ACME HTTP01 solver pods with expected resource configuration")
-
-			By("waiting for certificate to get ready")
-			err = waitForCertificateReadiness(ctx, secretName, ns.Name)
-			Expect(err).NotTo(HaveOccurred(), "timeout waiting for certificate to become ready")
 		})
 	})
 
--- a/test/e2e/issuer_acme_http01_test.go
+++ b/test/e2e/issuer_acme_http01_test.go
@@ -224,8 +224,16 @@ var _ = Describe("ACME Issuer HTTP01 solver", Label("Platform:Generic"), Ordered
 
 		It("should obtain a valid certificate", func() {
 
+			By("waiting for certificate to become ready")
+			err := waitForCertificateReadiness(ctx, secretName, ns.Name)
+			Expect(err).NotTo(HaveOccurred(), "timeout waiting for certificate to become ready")
+
 			By("checking TLS certificate contents")
-			err := wait.PollUntilContextTimeout(ctx, slowPollInterval, highTimeout, true, func(context.Context) (bool, error) {
+			err = verifyCertificate(ctx, secretName, ns.Name, ingressHost)
+			Expect(err).NotTo(HaveOccurred(), "certificate verification failed")
+
+			By("verifying live Ingress TLS endpoint")
+			err = wait.PollUntilContextTimeout(ctx, slowPollInterval, highTimeout, true, func(context.Context) (bool, error) {
 				secret, err := loader.KubeClient.CoreV1().Secrets(ns.Name).Get(ctx, secretName, metav1.GetOptions{})
 				if err != nil {
 					return false, nil // keep polling until the Secret exists
@@ -246,7 +254,7 @@ var _ = Describe("ACME Issuer HTTP01 solver", Label("Platform:Generic"), Ordered
 
 				return isHostCorrect && isNotExpired, nil
 			})
-			Expect(err).NotTo(HaveOccurred(), "timeout waiting for valid TLS certificate")
+			Expect(err).NotTo(HaveOccurred(), "timeout waiting for valid live Ingress TLS certificate")
 		})
 
 		It("should create solver pods with custom resource limits and requests", func() {
--- a/test/e2e/multiple_operands_test.go
+++ b/test/e2e/multiple_operands_test.go
@@ -1,0 +1,917 @@
+//go:build e2e
+// +build e2e
+
+// CM-786 IstioCSR Controller qualification (doc section "IstioCSR Controller specific tests"
+// through "Record about cert-manager Operator"): multi-operand install plus sanity and
+// qualification steps 5–14 that are automatable in e2e.
+package e2e
+
+import (
+	"context"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/url"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	cmmetav1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
+	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"golang.org/x/sync/errgroup"
+	"sigs.k8s.io/yaml"
+
+	testutils "github.com/openshift/cert-manager-operator/pkg/controller/istiocsr"
+	"github.com/openshift/cert-manager-operator/api/operator/v1alpha1"
+	"github.com/openshift/cert-manager-operator/test/library"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	istioCSRDeploymentName     = "cert-manager-istio-csr"
+	istioCSRServiceName        = "cert-manager-istio-csr"
+	istioCSRServiceAccountName = "cert-manager-istio-csr"
+
+	bothOperandsFeatureGates = "IstioCSR=true,TrustManager=true"
+
+	cm786QualificationLabelKey   = "cm-786-qualification"
+	cm786QualificationLabelValue = "true"
+
+	cm10TestDNSName = "multi-operand.cm10.test.example"
+	cm12TestDNSName = "multi-operand.cm12.invalid.example"
+
+	qualificationBundleName     = "cm786-ca-bundle-secret"
+	qualificationBundleSourceCM = "cm786-bundle-source"
+	qualificationBundleSourceKey = "ca-bundle.crt"
+	qualificationBundleTargetKey = "ca-bundle.crt"
+)
+
+type operandDeployment struct {
+	namespace string
+	name      string
+}
+
+var coreOperandDeployments = []operandDeployment{
+	{operandNamespace, certmanagerControllerDeployment},
+	{operandNamespace, certmanagerWebhookDeployment},
+	{operandNamespace, certmanagerCAinjectorDeployment},
+	{operandNamespace, trustManagerDeploymentName},
+}
+
+var _ = Describe("Multiple operands CM-786 qualification", Ordered, Label("Platform:Generic", "Feature:MultipleOperands", "TechPreview", "Skipped:MicroShift"), func() {
+	var (
+		ctx = context.Background()
+
+		clientset                        *kubernetes.Clientset
+		originalUnsupportedAddonFeatures string
+		originalOperatorLogLevel         string
+		istioNS                          *corev1.Namespace
+		qualificationBundlePEM           string
+	)
+
+	BeforeAll(func() {
+		var err error
+		clientset, err = kubernetes.NewForConfig(cfg)
+		Expect(err).NotTo(HaveOccurred())
+
+		if isMicroShiftCluster(ctx, clientset) {
+			Skip("MicroShift: OLM not available")
+		}
+
+		trustManagerRemoveStaleOperandServiceAccount(ctx, clientset)
+
+		originalUnsupportedAddonFeatures, err = getSubscriptionEnvVar(ctx, loader, "UNSUPPORTED_ADDON_FEATURES")
+		Expect(err).NotTo(HaveOccurred())
+
+		originalOperatorLogLevel, err = getSubscriptionEnvVar(ctx, loader, "OPERATOR_LOG_LEVEL")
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: test/e2e/issuer_acme_http01_test.go [hunk 1]
```diff
@@ -286,10 +286,6 @@ var _ = Describe("ACME Issuer HTTP01 solver", Label("Platform:Generic"), Ordered
 				return false, nil // No matching pods yet, keep waiting
 			})
 			Expect(err).NotTo(HaveOccurred(), "should find ACME HTTP01 solver pods with expected resource configuration")
-
-			By("waiting for certificate to get ready")
```
bharath-b-rh, line 290: Is this removed to avoid redundancy with `By("checking TLS certificate contents")`? If so, shouldn't this be moved to the `By` block to verify readiness before validating the individual properties in Certificate?

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+)
+
+const (
+	istioCSRDeploymentName     = "cert-manager-istio-csr"
+	istioCSRServiceName        = "cert-manager-istio-csr"
+	istioCSRServiceAccountName = "cert-manager-istio-csr"
+
+	bothOperandsFeatureGates = "IstioCSR=true,TrustManager=true"
+
+	cm786QualificationLabelKey   = "cm-786-qualification"
```
bharath-b-rh: Can we remove the identifiers? The CM-786 itself is very generic tracker created for release, and specific to the test cases added here.
Please remove at other places as well, like cm10, cm786 etc..

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+	BeforeAll(func() {
+		var err error
+		clientset, err = kubernetes.NewForConfig(cfg)
+		Expect(err).NotTo(HaveOccurred())
+
+		if isMicroShiftCluster(ctx, clientset) {
+			Skip("MicroShift: OLM not available")
+		}
+
+		trustManagerRemoveStaleOperandServiceAccount(ctx, clientset)
```
bharath-b-rh: What could cause the stale serviceaccount, just that? No other resources of trust-manager needs to be cleaned up?
I think the more apt change would be to clean it up where it's getting created, in `AfterAll` block.

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+
+		originalUnsupportedAddonFeatures, err = getSubscriptionEnvVar(ctx, loader, "UNSUPPORTED_ADDON_FEATURES")
+		Expect(err).NotTo(HaveOccurred())
+
+		originalOperatorLogLevel, err = getSubscriptionEnvVar(ctx, loader, "OPERATOR_LOG_LEVEL")
+		Expect(err).NotTo(HaveOccurred())
+
+		By("CM-786 step 8: enable IstioCSR and TrustManager feature gates")
```
bharath-b-rh: Please remove `step 8` like identifiers, these are coming from the doc I guess.

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+		WithTrustNamespace(trustManagerNamespace)
+}
+
+func installOperandsConcurrently(ctx context.Context, trustManager *v1alpha1.TrustManager, istioCSR *v1alpha1.IstioCSR) {
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	var tmErr, istioErr error
+
+	wg.Add(2)
+	go func() {
```
bharath-b-rh, line 273: Do we really need routines here? The Create API itself is a blocking call. Is it to create a concurrent request to respective controllers?

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+	close(start)
+	wg.Wait()
+
+	Expect(tmErr).NotTo(HaveOccurred())
+	Expect(istioErr).NotTo(HaveOccurred())
+}
+
+func waitForOperandsReadyConcurrently(ctx context.Context, clientset *kubernetes.Clientset, istioNamespace string) {
+	var eg errgroup.Group
+	eg.Go(func() error {
```
bharath-b-rh, line 293: The concurrency in previous method makes sense, but here I am unable to understand. Do we need to poll simultaneously? These are independent resources, a race condition or deadlock wouldn't be expected, AFAIU.
All said, I just wanted to understand, there is no harm in having it.

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+	}
+
+	return strings.TrimSpace(b.String())
+}
+
+func verifyAllOperandDeploymentsReady(ctx context.Context, clientset *kubernetes.Clientset, istioNamespace string) {
+	deployments := append([]operandDeployment{}, coreOperandDeployments...)
+	deployments = append(deployments, operandDeployment{istioNamespace, istioCSRDeploymentName})
+
+	for _, dep := range deployments {
```
bharath-b-rh, line 356: Why not put these into routines, so that it happens paralleley and should get completed faster I guess.

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+
+	crs, err := clientset.RbacV1().ClusterRoles().List(ctx, metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/name=cert-manager-istio-csr",
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(crs.Items).NotTo(BeEmpty())
+	verifyOperatorManagedLabels(crs.Items[0].Labels, "cert-manager-istio-csr")
+}
+
+func verifyTrustManagerQualificationConfig(ctx context.Context, clientset *kubernetes.Clientset) {
```
bharath-b-rh: Could we improve the naming? Something like `assertTrustManagerCRConfigPropagation`

test/e2e/utils_test.go
```diff
@@ -2085,15 +2097,15 @@ func formatVaultPodsStatus(pods []corev1.Pod) string {
 			case cs.State.Running != nil:
 				b.WriteString(" state=running")
 			case cs.State.Waiting != nil:
-				fmt.Fprintf(&b, " state=waiting reason=%s message=%q", cs.State.Waiting.Reason, cs.State.Waiting.Message)
```
bharath-b-rh, line 2088: Isn't message helpful during debugging?

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,917 @@
...
+
+		It("CM-786 step 13: IstioCSR API updates reconcile successfully", func() {
+			runCM786IstioCSRSpecUpdateTest(ctx, istioNS.Name)
+		})
+
+		It("CM-786 step 14: IstioCSR gRPC CreateCertificate returns a cert chain", func() {
+			runCM786IstioCSRGRPCCertificateTest(ctx, clientset, istioNS)
+		})
+
+		It("CM-786 steps 6-7: Service Mesh uses istio-csr when Service Mesh is installed", func() {
```
bharath-b-rh: I thought we are installing the ServiceMesh as well, is it not the case? Are we only relying on the sample gRPC requests for validating functionality?

test/e2e/issuer_acme_http01_test.go
arun717: Adding the check for ==> By("waiting for certificate to become ready") before
By("checking TLS certificate contents")

test/e2e/multiple_operands_test.go
arun717: yes sure. This got included since we had used the qe-e2e-tool which first generates the plan in a doc and then as a next step starts mapping in code while implementing.

test/e2e/multiple_operands_test.go
arun717: To answer what could cause a stale service account ?
>A prior e2e run created TrustManager, then deleted only the CR and not the operand ServiceAccount.

Moved cleanup from defensive BeforeAll to teardown where the TrustManager CR is deleted.

test/e2e/multiple_operands_test.go
arun717: done.

test/e2e/multiple_operands_test.go
arun717: Yes the goroutines are intentional. They are used since the test is checking concurrent install of both operands istio csr and trust manager and so that both controllers reconcile in parallel. I added more comments around this code block for better understanding.

test/e2e/multiple_operands_test.go
arun717: Good question Bharath. So, concurrent polling here isn’t about avoiding races or deadlocks. TrustManager and IstioCSR are independent and their sequential polls would work as well with no functional impact.

The code is written with the intention to poll in parallel mainly to cap wait time at ~one multiOperandReadyTimeout (20m) instead of two in series with respect to a slow CI.

test/e2e/multiple_operands_test.go
arun717: Done. Switched verifyAllOperandDeploymentsReady to parallel polling via errgroup (same pattern as waitForOperandsReadyConcurrently). Same assertions/timeouts which caps wall-clock wait at one lowTimeout instead of sequential checks.

test/e2e/multiple_operands_test.go
arun717: Done.

test/e2e/utils_test.go
arun717: Restored the diagnostic message fields in func formatVaultPodsStatus. I don't remember the exact reason why this was removed. Maybe the removal was unintentional !!!

test/e2e/multiple_operands_test.go
arun717: Correct  we are not installing Service Mesh in this PR. The suite installs TrustManager and IstioCSR operands concurrently and validates their health, cert-manager behavior, TrustManager config etc.

Yes ServiceMesh is not installed, gRPC validates the istio-csr operand API directly. 
The Service Mesh spec is a lightweight conditional wiring check that skips when mesh is absent.

I can create a separate PR later if we need to install OSSM and add a real mesh check.

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,930 @@
...
+	logStream, err := req.Stream(ctx)
+	Expect(err).NotTo(HaveOccurred())
+	defer logStream.Close()
+
+	logData, err := io.ReadAll(logStream)
+	Expect(err).NotTo(HaveOccurred())
+
+	var entry LogEntry
+	Expect(json.Unmarshal(logData, &entry)).To(Succeed())
+	Expect(entry.CertChain).NotTo(BeEmpty())
```
bharath-b-rh: Plain json unmarshal could cause flakiness when there are incomplete log messages. Use `parseGRPCurlLogEntry()`, used in other tests.
```suggestion
	entry, err := parseGRPCurlLogEntry(logData)
	Expect(err).Should(BeNil())
	Expect(entry.CertChain).ShouldNot(BeEmpty())
```

test/e2e/multiple_operands_test.go
```diff
@@ -0,0 +1,930 @@
...
+				"UNSUPPORTED_ADDON_FEATURES", originalUnsupportedAddonFeatures, lowTimeout)
+		}
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	Context("with operands installed concurrently", Ordered, func() {
+		BeforeAll(func() {
+			var err error
+
+			istioNS, err = loader.CreateTestingNS("multi-operand-istio", true)
```
bharath-b-rh: don't we need this to be dynamic, like in everyother testcase?
```suggestion
			istioNS, err = loader.CreateTestingNS("multi-operand-istio", false)
```

test/e2e/multiple_operands_test.go
bharath-b-rh: Yeah, sure please create a tracker for the same.

test/e2e/multiple_operands_test.go
arun717: done.

test/e2e/multiple_operands_test.go
arun717: yes, edited to false for dynamic name.

bharath-b-rh (approved):
/lgtm
/approve

@arun717 Please create a tracker for the ServiceMesh integration.

bharath-b-rh:
No user facing changes, hence adding below labels 
/label px-approved 
/label docs-approved 

Based on the CI jobs result, adding below label 
/label qe-approved

test/e2e/multiple_operands_test.go
arun717: created
https://redhat.atlassian.net/browse/CM-1224

PR #420, opened 2026-05-06
CM-1040: Extract shared ApplyResource helper and migrate istiocsr to SSA

sebrandon1:
## Summary

- Add `common.ApplyResource[T]()` in `pkg/controller/common/applier.go` — a type-safe SSA reconciliation function with pluggable drift-detection callbacks and standardized event recording
- Migrate trustmanager's 11 `createOrApply` methods to delegate to the shared helper
- Migrate istiocsr's 8 simple `createOrApply` methods from Create/UpdateWithRetry to SSA via the shared helper
- Handle RoleBinding roleRef immutability: when the roleRef changes, the binding is deleted and recreated
- Remove the `istioCSRCreateRecon` parameter and its "ResourceAlreadyExists" warning events — SSA is declarative and idempotent
- Net -212 lines of duplicated reconciliation boilerplate

## Related PRs

This is part of an ordered refactoring series:

| PR | Title | Status |
|----|-------|--------|
| [openshift/cert-manager-operator#418](https://github.com/openshift/cert-manager-operator/pull/418) | [CM-1038](https://redhat.atlassian.net/browse/CM-1038): Replace istiocsr type-specific decoders with generic DecodeObjBytes | Open |
| [openshift/cert-manager-operator#419](https://github.com/openshift/cert-manager-operator/pull/419) | [CM-1039](https://redhat.atlassian.net/browse/CM-1039): Thread context.Context from Reconcile() through controller helpers | Open |
| [openshift/cert-manager-operator#438](https://github.com/openshift/cert-manager-operator/pull/438) | [CM-1113](https://redhat.atlassian.net/browse/CM-1113): Replace unsafe.Pointer casts with Kubernetes conversion functions | Open |
| [openshift/cert-manager-operator#417](https://github.com/openshift/cert-manager-operator/pull/417) | [CM-1114](https://redhat.atlassian.net/browse/CM-1114): Add health probes and richer status conditions | Open |
| [openshift/cert-manager-operator#461](https://github.com/openshift/cert-manager-operator/pull/461) | [CNF-26153](https://redhat.atlassian.net/browse/CNF-26153): Add unit tests across codebase to close coverage gaps | Open |

## Jira

- [CM-1040](https://issues.redhat.com/browse/CM-1040) — Extract shared ApplyResource helper and migrate istiocsr to SSA

## Test Plan

- [x] New `applier_test.go` with unit tests for ApplyResource (create, no-op, update, Exists error, Patch error, error classification)
[...]

PR #419, opened 2026-05-06
CM-1039: Thread context.Context from Reconcile() through controller helpers

sebrandon1:
## Summary

- Removes the stored `ctx context.Context` field from both `istiocsr` and `trustmanager` Reconciler structs
- Threads context from `Reconcile(ctx context.Context, ...)` through every helper method call chain (37 files, ~90 call sites)
- Tests pass `context.Background()` explicitly (semantically correct for test entrypoints)
- Removes a redundant `sort.Strings` call after `MergeContainerArgs` which already sorts its output

## Motivation

Both controllers stored a `context.Context` initialized once to `context.Background()` in `New()`. The `Reconcile()` method receives a proper request-scoped context from controller-runtime, but all helper methods used the stale struct field. This defeats cancellation and deadline propagation from the framework.

This supersedes #242, which attempted to standardize context usage by replacing `context.Background()` with `context.TODO()` everywhere -- semantically backwards since `context.TODO()` signals "placeholder, haven't decided yet" while `context.Background()` is correct for top-level contexts.

## Related PRs

| PR | Title | Status |
|----|-------|--------|
| [openshift/cert-manager-operator#418](https://github.com/openshift/cert-manager-operator/pull/418) | [CM-1038](https://redhat.atlassian.net/browse/CM-1038): Replace istiocsr type-specific decoders with generic DecodeObjBytes | Open |
| [openshift/cert-manager-operator#420](https://github.com/openshift/cert-manager-operator/pull/420) | [CM-1040](https://redhat.atlassian.net/browse/CM-1040): Extract shared ApplyResource helper and migrate istiocsr to SSA | Open |
| [openshift/cert-manager-operator#438](https://github.com/openshift/cert-manager-operator/pull/438) | [CM-1113](https://redhat.atlassian.net/browse/CM-1113): Replace unsafe.Pointer casts with Kubernetes conversion functions | Open |
| [openshift/cert-manager-operator#417](https://github.com/openshift/cert-manager-operator/pull/417) | [CM-1114](https://redhat.atlassian.net/browse/CM-1114): Add health probes and richer status conditions | Open |
| [openshift/cert-manager-operator#461](https://github.com/openshift/cert-manager-operator/pull/461) | [CNF-26153](https://redhat.atlassian.net/browse/CNF-26153): Add unit tests across codebase to close coverage gaps | Open |

## Jira

- [CM-1039](https://issues.redhat.com/browse/CM-1039) — Thread context.Context from Reconcile() through controller helpers

## Test Plan

[...]

PR #438, opened 2026-06-10
CM-1113: Replace unsafe.Pointer casts with Kubernetes conversion functions

sebrandon1:
## Summary

- Replaces all 8 raw `unsafe.Pointer` casts between `corev1.*` and `core.*` types with Kubernetes auto-generated conversion functions from `k8s.io/kubernetes/pkg/apis/core/v1`
- Removes `"unsafe"` import from 3 files
- Adds `ToCoreTolerations` and `ToV1Tolerations` helpers in `pkg/controller/common/validation.go` to avoid repeating the conversion loop pattern

### Why

The operator uses Kubernetes internal validation functions (`corevalidation.ValidateTolerations`, etc.) which require internal `core.*` types, but the public API uses `corev1.*` types. Previously, this was bridged with raw `unsafe.Pointer` casts like:

```go
convTolerations := *(*[]core.Toleration)(unsafe.Pointer(&tolerations))
```

While the types are memory-layout compatible, raw unsafe casts are fragile and bypass Go's type system. Kubernetes already provides auto-generated conversion functions that handle this properly. The conversion functions still use `unsafe.Pointer` internally, but the cast is encapsulated, upstream-maintained, and follows the official Kubernetes conversion pattern.

## Related PRs

| PR | Title | Status |
|----|-------|--------|
| [openshift/cert-manager-operator#419](https://github.com/openshift/cert-manager-operator/pull/419) | [CM-1039](https://redhat.atlassian.net/browse/CM-1039): Thread context.Context from Reconcile() through controller helpers | Open |
| [openshift/cert-manager-operator#420](https://github.com/openshift/cert-manager-operator/pull/420) | [CM-1040](https://redhat.atlassian.net/browse/CM-1040): Extract shared ApplyResource helper and migrate istiocsr to SSA | Open |
| [openshift/cert-manager-operator#461](https://github.com/openshift/cert-manager-operator/pull/461) | [CNF-26153](https://redhat.atlassian.net/browse/CNF-26153): Add unit tests across codebase to close coverage gaps | Open |

## Jira

- [CM-1113](https://issues.redhat.com/browse/CM-1113) — Replace unsafe.Pointer casts with Kubernetes conversion functions

## Test Plan

- [x] `go build ./...` compiles cleanly
- [x] `make test` passes
- [x] `make lint` introduces no new issues
- [ ] CI e2e tests pass

### DIFF
--- a/pkg/controller/common/validation.go
+++ b/pkg/controller/common/validation.go
@@ -1,13 +1,12 @@
 package common
 
 import (
-	"unsafe"
-
 	corev1 "k8s.io/api/core/v1"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/kubernetes/pkg/apis/core"
+	corev1conversion "k8s.io/kubernetes/pkg/apis/core/v1"
 	corevalidation "k8s.io/kubernetes/pkg/apis/core/validation"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controller/common/validation.go
```diff
@@ -20,25 +19,23 @@ func ValidateNodeSelectorConfig(nodeSelector map[string]string, fldPath *field.P
 // ValidateTolerationsConfig validates the Tolerations configuration using
 // the Kubernetes core toleration validation rules.
 func ValidateTolerationsConfig(tolerations []corev1.Toleration, fldPath *field.Path) error {
-	// convert corev1.Tolerations to core.Tolerations, required for validation.
-	convTolerations := *(*[]core.Toleration)(unsafe.Pointer(&tolerations))
-	return corevalidation.ValidateTolerations(convTolerations, fldPath.Child("tolerations")).ToAggregate()
+	return corevalidation.ValidateTolerations(ToCoreTolerations(tolerations), fldPath.Child("tolerations")).ToAggregate()
 }
```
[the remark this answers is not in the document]
sebrandon1: These conversion functions are auto-generated by Kubernetes' conversion-gen and always return nil — they just copy fields between structurally identical types. You can verify by reading the implementations in `vendor/k8s.io/kubernetes/pkg/apis/core/v1/zz_generated.conversion.go`:

```go
func autoConvert_v1_Toleration_To_core_Toleration(in *corev1.Toleration, out *core.Toleration, s conversion.Scope) error {
    out.Key = in.Key
    out.Operator = core.TolerationOperator(in.Operator)
    out.Value = in.Value
    out.Effect = core.TaintEffect(in.Effect)
    out.TolerationSeconds = (*int64)(unsafe.Pointer(in.TolerationSeconds))
    return nil  // always nil
}
```

Same pattern for `Convert_v1_Affinity_To_core_Affinity` and `Convert_v1_ResourceRequirements_To_core_ResourceRequirements` — all unconditionally return nil.

Adding `if err != nil` handling for errors that can never occur would add boilerplate for an impossible code path and obscure the actual logic. The `_ =` assignment explicitly acknowledges the return value while keeping the code clean.

sebrandon1:
@coderabbitai resume

pkg/controller/common/validation.go
```diff
@@ -20,25 +19,23 @@ func ValidateNodeSelectorConfig(nodeSelector map[string]string, fldPath *field.P
...
 
 // ValidateAffinityRules validates the Affinity configuration using
 // the Kubernetes core affinity validation rules.
 func ValidateAffinityRules(affinity *corev1.Affinity, fldPath *field.Path) error {
-	// convert corev1.Affinity to core.Affinity, required for validation.
-	convAffinity := (*core.Affinity)(unsafe.Pointer(affinity))
-	return validateAffinity(convAffinity, corevalidation.PodValidationOptions{}, fldPath.Child("affinity")).ToAggregate()
+	var convAffinity core.Affinity
+	_ = corev1conversion.Convert_v1_Affinity_To_core_Affinity(affinity, &convAffinity, nil)
+	return validateAffinity(&convAffinity, corevalidation.PodValidationOptions{}, fldPath.Child("affinity")).ToAggregate()
```
anandkuma77, line 41: ValidateAffinityRules will panic if called with a nil affinity pointer.

sebrandon1: Good catch — added a nil guard at the top of `ValidateAffinityRules` (returns nil early) plus a `TestValidateAffinityRules_Nil` test to cover it.

PR #439, opened 2026-06-15
CM-1033: Add performance tuning parameters to the supported args whitelist

chiragkyal:
## Description

Adds `--concurrent-workers`, `--kube-api-qps`, `--kube-api-burst`, and `--max-concurrent-challenges` to the cert-manager controller's supported overrideArgs whitelist, enabling users to tune cert-manager performance via the CertManager CR without relying on unsupportedConfigOverrides.

## Changes

- Add 4 performance-tuning flag constants to the controller args whitelist
- Add `validateBurstQPS()` that checks `burst >= qps` when both are specified
- Unit tests: acceptance of new flags + rejection when burst < qps
- E2E tests: valid flags propagate to deployment + burst < qps causes Degraded

---

## Summary by CodeRabbit

* **New Features**
  * Added validation for controller “performance tuning” override args, including concurrency and kube API QPS/burst settings.

* **Bug Fixes**
  * Prevents invalid combinations from being applied, including rejecting cases where kube-api-burst is less than kube-api-qps and ensuring values are positive and correctly formatted.

* **Tests**
  * Expanded unit tests for accepted and rejected tuning scenarios with exact error assertions.
  * Enhanced e2e tests to confirm valid tuning is applied while invalid tuning is not propagated and controller health remains accurate.

### DIFF
--- a/pkg/controller/certmanager/deployment_overrides_validation.go
+++ b/pkg/controller/certmanager/deployment_overrides_validation.go
@@ -137,8 +137,9 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
 // validatePerformanceArgs performs sanity checks on the performance tuning
 // arguments to catch invalid configurations.
 func validatePerformanceArgs(argMap map[string]string) error {
-	// Validate that numeric args are positive integers where applicable.
-	positiveIntArgs := []string{argConcurrentWorkers, argMaxConcurrentChallenges}
+	// Validate that integer args are positive.
+	positiveIntArgs := []string{argConcurrentWorkers, argMaxConcurrentChallenges, argKubeAPIBurst}
+	parsedInts := make(map[string]int)
 	for _, arg := range positiveIntArgs {
 		if valStr, ok := argMap[arg]; ok {
 			val, err := strconv.Atoi(valStr)
@@ -148,30 +149,29 @@ func validatePerformanceArgs(argMap map[string]string) error {
 			if val <= 0 {
 				return fmt.Errorf("validation failed: %s must be greater than 0, got %d", arg, val)
 			}
+			parsedInts[arg] = val
 		}
 	}
 
-	// Validate QPS and burst are numeric and positive.
-	positiveFloatArgs := []string{argKubeAPIQPS, argKubeAPIBurst}
-	parsedFloats := make(map[string]float64)
-	for _, arg := range positiveFloatArgs {
-		if valStr, ok := argMap[arg]; ok {
-			val, err := strconv.ParseFloat(valStr, 64)
-			if err != nil {
-				return fmt.Errorf("validation failed: %s value must be numeric, got %q", arg, valStr)
-			}
-			if val <= 0 {
-				return fmt.Errorf("validation failed: %s must be greater than 0, got %v", arg, val)
-			}
-			parsedFloats[arg] = val
+	// Validate QPS is a positive float32
+	var qps float32
+	var qpsOk bool
+	if qpsStr, ok := argMap[argKubeAPIQPS]; ok {
+		val, err := strconv.ParseFloat(qpsStr, 32)
+		if err != nil {
+			return fmt.Errorf("validation failed: %s value must be numeric, got %q", argKubeAPIQPS, qpsStr)
 		}
+		if val <= 0 {
+			return fmt.Errorf("validation failed: %s must be greater than 0, got %v", argKubeAPIQPS, val)
+		}
+		qps = float32(val)
+		qpsOk = true
 	}
 
 	// Validate burst >= qps when both are specified.
-	qps, qpsOk := parsedFloats[argKubeAPIQPS]
-	burst, burstOk := parsedFloats[argKubeAPIBurst]
-	if qpsOk && burstOk && burst < qps {
-		return fmt.Errorf("validation failed: --kube-api-burst (%v) must be >= --kube-api-qps (%v)", burst, qps)
+	burst, burstOk := parsedInts[argKubeAPIBurst]
+	if qpsOk && burstOk && float32(burst) < qps {
+		return fmt.Errorf("validation failed: --kube-api-burst (%d) must be >= --kube-api-qps (%v)", burst, qps)
 	}
 
 	return nil
--- a/pkg/controller/certmanager/deployment_overrides_validation.go
+++ b/pkg/controller/certmanager/deployment_overrides_validation.go
@@ -2,6 +2,7 @@ package certmanager
 
 import (
 	"fmt"
+	"strconv"
 	"unsafe"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -18,6 +19,13 @@ import (
 	"github.com/openshift/cert-manager-operator/api/operator/v1alpha1"
 	"github.com/openshift/cert-manager-operator/pkg/controller/common"
 	certmanagerinformer "github.com/openshift/cert-manager-operator/pkg/operator/informers/externalversions/operator/v1alpha1"
+)
+
+const (
+	argConcurrentWorkers       = "--concurrent-workers"
+	argKubeAPIQPS              = "--kube-api-qps"
+	argKubeAPIBurst            = "--kube-api-burst"
+	argMaxConcurrentChallenges = "--max-concurrent-challenges"
 )
 
 // withContainerArgsValidateHook validates the container args with those that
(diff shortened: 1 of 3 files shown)


pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -115,6 +134,25 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
 	}
 }
 
+// validateBurstQPS ensures that if both --kube-api-burst and --kube-api-qps are
+// specified, burst is not less than qps (cert-manager crashes if burst < qps).
+func validateBurstQPS(argMap map[string]string) error {
```
chiragkyal: This validation is present on the operand side 

https://github.com/openshift/jetstack-cert-manager/blob/635387ca4d661102c327b97d5497c46bc79fa860/internal/apis/config/controller/validation/validation.go#L56-L58

So, this validation is optional here

pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -20,6 +21,13 @@ import (
 	certmanagerinformer "github.com/openshift/cert-manager-operator/pkg/operator/informers/externalversions/operator/v1alpha1"
 )
 
+const (
+	argConcurrentWorkers     = "--concurrent-workers"
```
bharath-b-rh: Needs `gofmt` fix here.

pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -115,6 +133,25 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
...
+	qpsStr, qpsOk := argMap[argKubeAPIQPS]
+	burstStr, burstOk := argMap[argKubeAPIBurst]
+	if !qpsOk || !burstOk {
+		return nil
+	}
+	qps, errQ := strconv.ParseFloat(qpsStr, 64)
+	burst, errB := strconv.ParseFloat(burstStr, 64)
+	if errQ != nil || errB != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
bharath-b-rh: @chiragkyal I agree with the suggestion.

pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -20,6 +21,13 @@ import (
 	certmanagerinformer "github.com/openshift/cert-manager-operator/pkg/operator/informers/externalversions/operator/v1alpha1"
 )
 
+const (
+	argConcurrentWorkers     = "--concurrent-workers"
+	argKubeAPIQPS            = "--kube-api-qps"
```
bharath-b-rh: Are these constants necessary? They are not made use of in uts or e2es.

pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -65,6 +73,14 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
...
 		"--certificate-request-minimum-backoff-duration",
+		// The number of concurrent workers for each controller. (default 5)
+		argConcurrentWorkers,
+		// Maximum queries-per-second to the Kubernetes API server. (default 20)
+		argKubeAPIQPS,
+		// Maximum burst queries-per-second to the Kubernetes API server. Must be >= kube-api-qps. (default 50)
+		argKubeAPIBurst,
+		// Maximum number of challenges that can be scheduled as 'processing' at once. (default 60)
+		argMaxConcurrentChallenges,
 	}
```
bharath-b-rh, line 84: nit: It would be good to have some sanity checks for the args. It's not there for others, but good to start from these. The operator doesn't monitor the pod status once after deployment resource creation, and if any of these args have invalid values and causes `CrashLoopBackOff` error stays in that state until someone notices and fixes it. There is liveness probe in-place, so no downtime is expected though.

pkg/controller/certmanager/deployment_overrides_validation.go
bharath-b-rh: Still good to have, I think.

test/e2e/overrides_test.go
```diff
@@ -148,6 +153,27 @@ var _ = Describe("Overrides test", Ordered, Label("Platform:Generic"), func() {
 		})
 	})
 
+	Context("When adding kube-api-burst less than kube-api-qps", func() {
```
bharath-b-rh, line 177: Could we add a happy path test case too?

bharath-b-rh (commented):
LGTM, except for few suggestions.

test/e2e/overrides_test.go
chiragkyal: Added a new test. This tests the dedicated happy path for all 4 performance flags.

pkg/controller/certmanager/deployment_overrides_validation.go
chiragkyal: Added validation checks for other parameters, too.

pkg/controller/certmanager/deployment_overrides_validation.go
chiragkyal: Not in test files, but in the same file, they have been used more than once. Hence, I think a constant makes sense.

pkg/controller/certmanager/deployment_overrides_validation.go
chiragkyal: Fixed

pkg/controller/certmanager/deployment_overrides_validation.go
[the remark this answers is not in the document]
chiragkyal: Addressed

pkg/controller/certmanager/deployment_overrides_validation.go
```diff
@@ -115,6 +134,49 @@ func withContainerArgsValidateHook(certmanagerinformer certmanagerinformer.CertM
...
+				return fmt.Errorf("validation failed: %s value must be a positive integer, got %q", arg, valStr)
+			}
+			if val <= 0 {
+				return fmt.Errorf("validation failed: %s must be greater than 0, got %d", arg, val)
+			}
+		}
+	}
+
+	// Validate QPS and burst are numeric and positive.
+	positiveFloatArgs := []string{argKubeAPIQPS, argKubeAPIBurst}
```
bharath-b-rh: argKubeAPIBurst must be parsed as Int or Float? AFAIU it should be Int, since cert-manager also accepts it as Int.
Ref: https://github.com/cert-manager/cert-manager/blob/master/cmd/controller/app/options/options.go#L73

And also argKubeAPIQPS must be parsed as float32 and not float64, isn't it?
Ref: https://github.com/cert-manager/cert-manager/blob/master/cmd/controller/app/options/options.go#L72

bharath-b-rh:
cc: @snarayan-redhat for docs approval. Please help add release notes for the changes.

```
## New performance-tuning override arguments for the cert-manager controller (CM-1033)

With this release, the cert-manager Operator for Red Hat OpenShift supports configuring performance-tuning parameters for the cert-manager controller by using the `overrideArgs` field of the `CertManager` custom resource (CR). Previously, users had to rely on `spec.unsupportedConfigOverrides` to tune these settings.

You can now set the following arguments under `spec.controllerConfig.overrideArgs`:

- `--concurrent-workers`: The number of concurrent workers for each controller. The default value is `5`.
- `--kube-api-qps`: The maximum number of queries per second sent to the Kubernetes API server. The default value is `20`.
- `--kube-api-burst`: The maximum burst of queries per second sent to the Kubernetes API server. Must be greater than or equal to `--kube-api-qps`. The default value is `50`.
- `--max-concurrent-challenges`: The maximum number of ACME challenges that can be scheduled as processing at the same time. The default value is `60`.

The operator validates that `--kube-api-burst` is greater than or equal to `--kube-api-qps` when both values are set. If this constraint is not met, the operator sets the `Degraded` condition on the `CertManager` CR and does not apply the invalid configuration to the controller deployment.

For more information, see _Overridable arguments for the cert-manager components_.
```

pkg/controller/certmanager/deployment_overrides_validation.go
chiragkyal: Thanks for the catch. Updated!

chiragkyal: We need to cast burst to float32 during comparison `float32(burst) < qps`; upstream is also doing the same way. 

https://github.com/openshift/jetstack-cert-manager/blob/635387ca4d661102c327b97d5497c46bc79fa860/internal/apis/config/controller/validation/validation.go#L56

bharath-b-rh:
/lgtm

adding label based on the e2e cases result
/label qe-approved
/label px-approved

PR #447, opened 2026-06-25
NO-JIRA: e2e: fix label-filter shell quoting in test-e2e target

tashirj:
## Problem
`make test-e2e` fails immediately with:
  Syntax Error Parsing Label Filter
  Platform:
  Missing set operation.

The shell word-splits the unquoted $(E2E_GINKGO_LABEL_FILTER) expansion,
so Ginkgo only sees "Platform:" instead of the full expression.

## Fix
Wrap $(E2E_GINKGO_LABEL_FILTER) in single quotes at the call site.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test execution by making label-based test filtering more reliable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -283,7 +283,7 @@ test-e2e: test-e2e-wait-for-stable-state ## Run end-to-end tests.
 		-timeout $(E2E_TIMEOUT) \
 		-count 1 -v -p 1 \
 		-tags e2e -run "$(TEST)" . \
-		-ginkgo.label-filter=$(E2E_GINKGO_LABEL_FILTER)
+		-ginkgo.label-filter='$(E2E_GINKGO_LABEL_FILTER)'
 
 .PHONY: test-e2e-wait-for-stable-state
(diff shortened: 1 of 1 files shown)


tashirj:
@coderabbitai review

bharath-b-rh:
This is not a must required change. Instead the make variable `E2E_GINKGO_LABEL_FILTER` can be leveraged as required to set the label filter. The change is for users convenience.

`E2E_GINKGO_LABEL_FILTER='"Platform: xyz"'` can be used too.

/ok-to-test

Could you please add a valid jira ID in the title, if there is one? Else NO-JIRA can be used.

bharath-b-rh:
/retitle NO-JIRA: e2e: fix label-filter shell quoting in test-e2e target

bharath-b-rh:
override failing jobs, which is because of label parsing issue, addressed in https://github.com/openshift/release/pull/81499
/override e2e-operator

PR #449, opened 2026-06-26
CM-1223: Add extensive e2e automation for TLS profile test

chiragkyal:
## Summary

The cert-manager operator propagates the cluster-wide TLS security profile and adherence policy to cert-manager operand deployments. Existing e2e coverage only validated a single happy-path case, leaving profile types, adherence transitions, rollback, and override behavior untested.

## Change 

Rework the TLS profile e2e suite to add broad coverage across:

- All TLS profile types (Modern, Intermediate, Old, Custom) under strict adherence

- Legacy adherence behavior (no args injected)

- Rollback from strict to legacy adherence

- `unsupportedConfigOverrides` precedence over cluster-enforced TLS settings

- Live profile switching between profiles

## Summary by CodeRabbit

* **New Features**
  * Expanded TLS profile support across adherence modes, overrides, rollback, and live transitions.
  * Added validation for TLS settings and cipher configurations.

* **Bug Fixes**
  * Improved consistency checks during TLS profile changes.
  * Increased resilience when validating deployment health during temporary service interruptions.

* **Tests**
  * Expanded end-to-end coverage for TLS profiles, certificates, and cipher combinations.
  * Added configurable validation timeouts for more reliable health checks.

### DIFF
--- a/test/e2e/tls_profile_test.go
+++ b/test/e2e/tls_profile_test.go
@@ -5,67 +5,363 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	configapiv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cert-manager-operator/api/operator/v1alpha1"
 	"github.com/openshift/cert-manager-operator/pkg/tlsprofile"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+// tlsProfileTestDeployments lists every cert-manager operand that the TLS hook touches.
+var tlsProfileTestDeployments = []string{
+	certmanagerControllerDeployment,
+	certmanagerWebhookDeployment,
+	certmanagerCAinjectorDeployment,
+}
+
+// tlsProfileModernProfile is {"type":"Modern","modern":{}} — the OpenShift API union requires
+// the matching sub-object to be present alongside the type discriminator.
+var tlsProfileModernProfile = &configapiv1.TLSSecurityProfile{
+	Type:   configapiv1.TLSProfileModernType,
+	Modern: &configapiv1.ModernTLSProfile{},
+}
+
+var tlsProfileIntermediateProfile = &configapiv1.TLSSecurityProfile{
+	Type:         configapiv1.TLSProfileIntermediateType,
+	Intermediate: &configapiv1.IntermediateTLSProfile{},
+}
+
+var tlsProfileOldProfile = &configapiv1.TLSSecurityProfile{
+	Type: configapiv1.TLSProfileOldType,
+	Old:  &configapiv1.OldTLSProfile{},
+}
+
+// tlsProfileCustomProfile uses a small cipher list that satisfies the HTTP/2 requirement
+// (must include ECDHE-RSA-AES128-GCM-SHA256 or ECDHE-ECDSA-AES128-GCM-SHA256).
+var tlsProfileCustomProfile = &configapiv1.TLSSecurityProfile{
+	Type: configapiv1.TLSProfileCustomType,
+	Custom: &configapiv1.CustomTLSProfile{
+		TLSProfileSpec: configapiv1.TLSProfileSpec{
+			MinTLSVersion: configapiv1.VersionTLS12,
+			Ciphers: []string{
+				"ECDHE-RSA-AES128-GCM-SHA256",   // HTTP/2 required
+				"ECDHE-RSA-AES256-GCM-SHA384",
+				"ECDHE-ECDSA-AES256-GCM-SHA384",
+			},
+		},
+	},
+}
+
 var _ = Describe("Cluster TLS security profile", Label("Platform:Generic", "Feature:TLSProfile", "TechPreview"), Ordered, func() {
-	var ctx context.Context
+	var (
+		ctx      context.Context
+		original *apiserverTLSConfig
+	)
 
 	BeforeAll(func() {
 		ctx = context.Background()
+
+		By("saving original apiserver TLS configuration")
+		var err error
+		original, err = getClusterAPIServerTLSConfig(ctx)
+		if apierrors.IsNotFound(err) {
+			Skip("apiserver.config.openshift.io/cluster not available on this cluster")
+		}
+		Expect(err).NotTo(HaveOccurred(), "failed to read original apiserver TLS config")
+
+		By("verifying tlsAdherence field is available in this cluster's API schema")
+		err = updateClusterAPIServerTLSConfig(ctx, nil,
+			configapiv1.TLSAdherencePolicyLegacyAdheringComponentsOnly)
+		if isTLSAdherenceUnsupported(err) {
+			Skip(fmt.Sprintf("apiserver tlsAdherence field not available on this cluster: %v", err))
+		}
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterAll(func() {
+		By("restoring original apiserver TLS configuration")
+		Eventually(func() error {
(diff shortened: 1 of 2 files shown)


test/e2e/tls_profile_test.go
```diff
@@ -5,67 +5,363 @@ package e2e
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterAll(func() {
+		By("restoring original apiserver TLS configuration")
+		Eventually(func() error {
+			return restoreClusterAPIServerTLSConfig(ctx, original)
+		}, lowTimeout, fastPollInterval).Should(Succeed())
+
+		By("clearing any unsupportedConfigOverrides left by scenario 8")
```
bharath-b-rh:
```suggestion
		By("clearing unsupportedConfigOverrides left by any test scenarios")
```

test/e2e/tls_profile_test.go
```diff
@@ -5,67 +5,363 @@ package e2e
...
 	})
 
-	It("should configure operand container TLS args from apiserver cluster profile", func() {
-		original, err := getClusterAPIServerTLSConfig(ctx)
-		if apierrors.IsNotFound(err) {
-			Skip("apiserver.config.openshift.io/cluster is not available on this cluster")
+	// ─────────────────────────────────────────────────────────────────────────
+	// Scenario 1: LegacyAdheringComponentsOnly — operands must be untouched
+	// ─────────────────────────────────────────────────────────────────────────
+	It("S1: should not inject TLS args when tlsAdherence is LegacyAdheringComponentsOnly", func() {
```
arun717, line 111: I was just comparing scenario 1 and scenario 6. And looks like S6 is a superset of S1.
So if S6 passes (no args injected with Modern + Legacy), S1 should always pass too.
S1 only adds the nil profile case, which does not exercise a different branch today.
I think even if we drop scenario 1 in this case, we should be good. Let me know what do you think.

anandkuma77 (approved):

chiragkyal:
/cc @bharath-b-rh @mytreya-rh 
for labels

test/e2e/tls_profile_test.go
chiragkyal: Thanks for the review. I see the logic, but I think S1 is worth keeping for a few reasons:

They test different API-level guarantees, not just different code paths.

S1 asserts: "when no TLS profile has ever been set, cert-manager operands are untouched." This is the day-0 / fresh-install state of any cluster.

S6 asserts: "when a profile is explicitly set to Modern but adherence is Legacy, operands are still untouched."
Both pass through the same early-return today, but they protect against independently plausible regression e.g:

- A future change that adds "apply a default profile when tlsSecurityProfile is nil" — this would break S1 but leave S6 green, because S6 has an explicit non-nil profile.

There is also a subtle documentation value: S1 makes it explicit that cert-manager's TLS hook is opt-in — zero configuration means zero injection. S6 alone doesn't communicate that; it looks like the adherence policy is the only gate, whereas S1 shows the nil profile is also a valid no-op state.

Also since it's the first test in the Ordered suite, it also serves as a clean-slate baseline before all the mutation tests that follow. Happy to discuss further, but I'd prefer to keep both.

test/e2e/tls_profile_test.go
```diff
@@ -5,67 +5,363 @@ package e2e
...
+		cm, err := certmanageroperatorclient.OperatorV1alpha1().CertManagers().
+			Get(ctx, "cluster", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		updated := cm.DeepCopy()
+		if args == nil {
+			// Clear the whole unsupportedConfigOverrides field
+			updated.Spec.OperatorSpec.UnsupportedConfigOverrides = runtime.RawExtension{}
```
bharath-b-rh, line 348: nit: This method is specifically for webhook, but cleanup is done for all components.

test/e2e/tls_profile_test.go
```diff
@@ -5,67 +5,363 @@ package e2e
...
+			To(Succeed(), "webhook must update to VersionTLS13 after profile switch")
+
+		By("verifying cipher suite args are absent after switch to TLS 1.3")
+		Expect(verifyDeploymentArgs(k8sClientSet, certmanagerWebhookDeployment,
+			tlsprofile.CertManagerCipherSuiteArgKeys, false)).
+			To(Succeed(), "cipher suite args must be removed after switching to Modern/TLS 1.3")
+
+		By("verifying all operands match the Modern profile")
+		expectedSpec, err := tlsprofile.EffectiveSpec(tlsProfileModernProfile)
+		Expect(err).NotTo(HaveOccurred())
```
bharath-b-rh: Could we add some message here to make debugging easier. Same at other places too.

bharath-b-rh (commented):
All the tests added here are getting skipped. Please check https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cert-manager-operator/449/pull-ci-openshift-cert-manager-operator-master-e2e-operator-tech-preview/2070425619958599680/artifacts/e2e-operator-tech-preview/test/artifacts/junit.xml for the reason.

chiragkyal:
> All the tests added here are getting skipped. Please check https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cert-manager-operator/449/pull-ci-openshift-cert-manager-operator-master-e2e-operator-tech-preview/2070425619958599680/artifacts/e2e-operator-tech-preview/test/artifacts/junit.xml for the reason.

I see the issue. The techpreview jobs are running in a HyperShift Cluster, where the `apiserver.config.openshift.io/cluster` object is immutable from inside the guest cluster.

```
apiserver tlsAdherence field not available on this cluster: apiservers.config.openshift.io &#34;cluster&#34; is forbidden: ValidatingAdmissionPolicy &#39;config&#39; with binding &#39;config-binding&#39; denied request: This resource cannot be created, updated, or deleted. Please ask your administrator to modify the resource in the HostedCluster object
```

So to run these tests need would need a standalone OCP cluster with `TechPreviewNoUpgrade` FeatureSet enabled. 

@bharath-b-rh, should we migrate the job to a standalone OCP instead of HyperShift, or do we have some alternative approach here?

test/e2e/tls_profile_test.go
chiragkyal: Added a comment to explain the intention.

test/e2e/tls_profile_test.go
chiragkyal: Added. Thanks!

chiragkyal:
> > All the tests added here are getting skipped. Please check https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cert-manager-operator/449/pull-ci-openshift-cert-manager-operator-master-e2e-operator-tech-preview/2070425619958599680/artifacts/e2e-operator-tech-preview/test/artifacts/junit.xml for the reason.
> 
> I see the issue. The techpreview jobs are running in a HyperShift Cluster, where the `apiserver.config.openshift.io/cluster` object is immutable from inside the guest cluster.
> 
> ```
> apiserver tlsAdherence field not available on this cluster: apiservers.config.openshift.io &#34;cluster&#34; is forbidden: ValidatingAdmissionPolicy &#39;config&#39; with binding &#39;config-binding&#39; denied request: This resource cannot be created, updated, or deleted. Please ask your administrator to modify the resource in the HostedCluster object
> ```
> 
> So to run these tests need would need a standalone OCP cluster with `TechPreviewNoUpgrade` FeatureSet enabled.
> 
> @bharath-b-rh, should we migrate the job to a standalone OCP instead of HyperShift, or do we have some alternative approach here?

The issue got fixed as part of https://github.com/openshift/release/pull/83634, which means the TLS tests will only run as part of `tls-scanner` job and will not get scheduled on HyperShift TechPreview profile.

arun717:
I think tls-scanner ci job will pass once this PR gets merged. https://github.com/openshift/cert-manager-operator/pull/466

bharath-b-rh (approved):
/lgtm

/hold rebase once https://github.com/openshift/cert-manager-operator/pull/466 is merged, for clear CI results.

PR #452, opened 2026-07-01
NO-JIRA: Fix Vault e2e Helm installer pod PSA seccomp violation

bharath-b-rh:
## Summary
The Vault Issuer e2e tests intermittently fail in `BeforeEach` when creating the `vault-installer` Helm pod on clusters that enforce the Pod Security Admission `restricted` profile. The pod spec was missing a required seccomp profile.

This change sets `seccompProfile.type: RuntimeDefault` on both the pod and container security contexts for the Helm installer pod in `setupVaultServer()`, matching the pattern used by the operator's own deployments.

## Problem
On clusters enforcing `restricted:latest` PSA, pod creation fails with:

```
pods "vault-installer" is forbidden: violates PodSecurity "restricted:latest": seccompProfile (pod or container "helm" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

The failure is flaky because PSA enforcement depends on cluster configuration. Tests pass on clusters with permissive or audit-only defaults, and fail on clusters (including many OpenShift and generic CI environments) that strictly enforce the restricted profile.

## Summary by CodeRabbit

* **Chores**
  * Updated container security settings to use the default seccomp profile in setup routines, aligning with current platform hardening practices.

### DIFF
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -2199,6 +2199,11 @@ func setupVaultServer(ctx context.Context, cfg *rest.Config, loader library.Dyna
 		Spec: corev1.PodSpec{
 			ServiceAccountName: serviceAccountName,
 			RestartPolicy:      corev1.RestartPolicyNever,
+			SecurityContext: &corev1.PodSecurityContext{
+				SeccompProfile: &corev1.SeccompProfile{
+					Type: corev1.SeccompProfileTypeRuntimeDefault,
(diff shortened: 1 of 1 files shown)


bharath-b-rh:
/cherrypick cert-manager-1.20

no user facing changes hence adding other labels
/label docs-approved
/label qe-approved
/label px-approved

PR #424, opened 2026-05-13
OCPBUGS-85579: Add Console capability annotation to ConsoleYAMLSample resources

sebrandon1:
## Summary

Add `capability.openshift.io/name: Console` annotation to the three `ConsoleYAMLSample` resources so OLM skips them on clusters where the Console capability is disabled.

## Problem

On clusters without the Console capability (e.g., RAN RDS deployments), the cert-manager operator install fails with:

```
api-server resource not found installing ConsoleYAMLSample cert-manager-acme-issuer-sample:
GroupVersionKind console.openshift.io/v1, Kind=ConsoleYAMLSample not found on the cluster.
```

PR [#355](https://github.com/openshift/cert-manager-operator/pull/355) added Console resources (QuickStart + YAML samples) to the operator bundle. The `ConsoleQuickStart` correctly includes the `capability.openshift.io/name: Console` annotation, but the three `ConsoleYAMLSample` resources were missed.

## Fix

Add `capability.openshift.io/name: Console` annotation to:
- `config/console/cert-manager-acme-issuer-sample.yaml`
- `config/console/cert-manager-certificate-sample.yaml`
- `config/console/cert-manager-issuer-sample.yaml`
- Corresponding `bundle/manifests/` files

This tells OLM to skip these optional UI resources when Console is not available, allowing the operator to install successfully on Console-less clusters.

## Testing

- Verified the annotated resources apply correctly on a Console-enabled cluster (OCP 4.22)
- Annotation format matches the existing `ConsoleQuickStart` resource from PR #355
- The `ConsoleYAMLSample` resources are cosmetic (YAML examples in the web console sidebar) and not required for cert-manager functionality

Jira: [CNF-23785](https://redhat.atlassian.net/browse/CNF-23785)

## Summary by CodeRabbit

* **Chores**
  * Added a Console capability metadata annotation to cert-manager console YAML samples across bundled and config manifests.
  * Applies to ACME issuer, certificate, and issuer sample manifests so console samples consistently include the capability.openshift.io/name: Console annotation.

Issue #249, opened 2025-03-04
CertManager


vtamayor:
Hi,

 
I wanto to know how to useCertManager,  to automate the generation of certificates for the entire Openshift cluster.
 
Do you have the documentation on how to get the information required so that you can share it with me?

Below the information that we need to obtain to integrate with Openshift
  
Secret: Secret within the kafka namespaces with API Token or Venafi Credentials.
Zone: Route of the policy zone from which the certificates will be issued.
URL: Venafi URL address to connect from Cert Manager.
 
Thanks and regards

lunarwhite:
Upstream Venafi issuer doc would be a great start: https://cert-manager.io/docs/configuration/venafi/

Regarding RH's Operator specific capabilities, you could ref https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/security_and_compliance/cert-manager-operator-for-red-hat-openshift#cert-manager-about_cert-manager-operator-about

Issue #148, opened 2023-09-20
Why is the cert-manager-operator for Red Hat not AllNamespaces


bitscuit:
During tech preview, I was able to install cert-manager-operator in AllNamespaces mode. When was this support removed and why?

The CNCF cert-manager-operator is only installable in AllNamespaces mode, so why does the Red Hat one deviate from this?

![image](https://github.com/openshift/cert-manager-operator/assets/14062475/cf26f7aa-e3de-4bdb-aabe-e98950b354b8)

Additionally, cert-manager is a cluster wide service, and there is a limitation that there cannot be 2 active cert-manager-controllers on a cluster simultaneously since cert-manager never implemented this functionality properly https://github.com/cert-manager/cert-manager/issues/2525.

So with Red Hat cert-manager not restricting the install mode to only AllNamespaces, isn't there a risk that different users on a cluster accidentally install Red Hat cert-manager multiple times, leading to multiple active cert-manager-controllers?

leifmadsen:
I actually have run into this today. When you install in multiple namespaces, the Deployment.app for the operator conflicts, and is constantly updating.

![image](https://github.com/openshift/cert-manager-operator/assets/768751/e649a14c-c680-4c9b-89d9-f3a37a16d406)

leifmadsen:
While it seems most currently-developed Operators are moving to AllNamespaces mode, it is quite disruptive to change anything like this mid-channel version.  Documenting why this namespace method was used could be useful, but otherwise I'm happy to see this close.

mikekaczmarski:
The controller deployed by RH Cert Manager operator has cluster-wide permissions and visibility, as does the upstream CNCF/Jetstack Cert-Manager on which this is based.
Not installing the Cert-Manager in AllNamespaces mode does not change this.  Allowing installation in anything but AllNamespaces mode is misleading.  This is the essence of the bug

mikekaczmarski:
On further review of this, I believe that the desire on behalf of Red Hat was to install the operator in a namespace other than openshift-operators even though it is cluster scoped (essentially AllNamespace Mode).  Taking the default, the operator is installed in the cert-manager-operator namespace with its deployed resources running in the cert-manager namespace.   So this operator's payload does indeed have cluster scoped visibility and control (look at its clusterrolebindings).
[...]

PR #455, opened 2026-07-08
OCPBUGS-85579: Move Console resources from bundle to runtime with capability detection

sebrandon1:
## Problem

cert-manager-operator fails on consoleless clusters (RAN/RDS/SNO) when OLM tries to install Console resources (ConsoleYAMLSample, ConsoleQuickStart):
```
GroupVersionKind console.openshift.io/v1, Kind=ConsoleYAMLSample not found
```

**Why PR #424's approach didn't work:**
The `capability.openshift.io/name: Console` annotation is only processed by CVO for core platform components — **not by OLM** for third-party operator bundles.

## Solution

Move Console resource creation from the OLM bundle to operator runtime, following the pattern from [cluster-monitoring-operator #2011](https://github.com/openshift/cluster-monitoring-operator/pull/2011).

**How it works:**
1. Operator checks `ClusterVersion.status.capabilities.enabledCapabilities` for Console capability
2. Console enabled → creates resources at runtime
3. Console disabled → skips silently

## Changes

- **Removed** Console manifests from `bundle/manifests/` (OLM no longer creates them)
- **Added** `pkg/controller/certmanager/console_resources.go` — runtime Console capability detection
- **Added** Console YAML files to `bindata/console/` for runtime embedding
- **Added** RBAC for Console resources and ClusterVersion access

## Behavior

| Cluster Type | Result |
|-------------|--------|
| Consoleless (RAN/RDS/SNO) | ✅ InstallPlan succeeds, Console resources not created |
| Console-enabled (standard) | ✅ InstallPlan succeeds, Console resources created at runtime |

## Testing

Verified on OCP 4.22:
- Console-enabled: InstallPlan Complete, no regression
- Simulated consoleless: InstallPlan Complete, no GVK errors
- Unit tests: 13 new tests covering capability detection and sync behavior

## References

- **Supersedes:** #424
- **Pattern reference:** https://github.com/openshift/cluster-monitoring-operator/pull/2011
- **OLM optional manifests (never implemented):** https://github.com/operator-framework/operator-lifecycle-manager/issues/2375

<details>
<summary><b>Other operators that solved this same problem</b></summary>

### Runtime capability/API detection (the pattern this PR follows):

| Operator | PR | Console Resource | Approach |
|----------|-----|-----------------|----------|
| cluster-monitoring-operator | [#2011](https://github.com/openshift/cluster-monitoring-operator/pull/2011) | ConsolePlugin | ClusterVersion.status.capabilities check (OCPBUGS-14922) |
| cluster-authentication-operator | [#589](https://github.com/openshift/cluster-authentication-operator/pull/589) | Operator restart | Watches ClusterVersion for Console capability (OCPBUGS-3440) |
| oadp-operator | [#2170](https://github.com/openshift/oadp-operator/pull/2170) | ConsoleCLIDownload | REST mapper probe / meta.IsNoMatchError ([OADP-7868](https://redhat.atlassian.net/browse/OADP-7868)) |
| gitops-operator | [#1172](https://github.com/redhat-developer/gitops-operator/pull/1172) | ConsolePlugin backend | config.openshift.io discovery check ([GITOPS-9943](https://redhat.atlassian.net/browse/GITOPS-9943)) |
| tektoncd/operator (Pipelines) | [#220](https://github.com/tektoncd/operator/pull/220) | ConsoleYAMLSample, ConsoleQuickStart | Runtime CRD existence check |
| hyperconverged-cluster-operator | [controller-runtime #2456](https://github.com/kubernetes-sigs/controller-runtime/issues/2456) | ConsoleQuickStart, ConsolePlugin | Fixed upstream — lazy cache type validation |
| stolostron/hypershift-addon-operator | [#202](https://github.com/stolostron/hypershift-addon-operator/pull/202) | ConsoleCLIDownload | Skip on consoleless (ACM-3863) |
| assisted-installer | [#574](https://github.com/openshift/assisted-installer/pull/574) | Installer hang | Skip waiting for console operator ([MGMT-12471](https://redhat.atlassian.net/browse/MGMT-12471)) |

### CVO annotation approach (works for CVO-managed components only, NOT for OLM operators):

| Operator | PR | Console Resource | Notes |
|----------|-----|-----------------|-------|
| console-operator | [#665](https://github.com/openshift/console-operator/pull/665), [#992](https://github.com/openshift/console-operator/pull/992), [#1082](https://github.com/openshift/console-operator/pull/1082) | All Console CRDs | Self-annotation ([CONSOLE-3220](https://redhat.atlassian.net/browse/CONSOLE-3220)) |
| cluster-monitoring-operator | [#2254](https://github.com/openshift/cluster-monitoring-operator/pull/2254) | Dashboard ConfigMaps | Required renaming manifests to avoid CVO force-enable (OCPBUGS-18326) |
| cluster-kube-apiserver-operator | [#1542](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1542), [#1565](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1565) | Dashboard ConfigMaps | Same rename workaround needed (OCPBUGS-18247, OCPBUGS-20331) |
| external-secrets-operator | [#148](https://github.com/openshift/external-secrets-operator/pull/148) | ConsoleYAMLSample | CVO annotation (same bug as cert-manager — ineffective for OLM) |
| secrets-store-csi-driver-operator | [#95](https://github.com/openshift/secrets-store-csi-driver-operator/pull/95) | ConsoleQuickStart, ConsoleYAMLSample | CVO annotation ([SSCSI-235](https://redhat.atlassian.net/browse/SSCSI-235)) |
| cluster-version-operator | [#1388](https://github.com/openshift/cluster-version-operator/pull/1388) | ConsolePlugin | CVO annotation ([OTA-1956](https://redhat.atlassian.net/browse/OTA-1956), TechPreview WIP) |

### Annotation removed / Console resources deleted entirely:

| Operator | PR | Notes |
|----------|-----|-------|
| lightspeed-agentic-operator | [#249](https://github.com/openshift/lightspeed-agentic-operator/pull/249) | Removed ConsolePlugin deployment entirely ([OLS-3350](https://redhat.atlassian.net/browse/OLS-3350)) |

### Test framework fixes:

| Component | PR | Notes |
|-----------|-----|-------|
| openshift/origin | [#27439](https://github.com/openshift/origin/pull/27439) | Console disruption monitoring conditional ([TRT-573](https://redhat.atlassian.net/browse/TRT-573)) |
| openshift/origin | [#27681](https://github.com/openshift/origin/pull/27681) | Console RBAC expectations conditional |
| openshift/assisted-service | [#5022](https://github.com/openshift/assisted-service/pull/5022) | Cluster deployment completion check ([MGMT-13941](https://redhat.atlassian.net/browse/MGMT-13941)) |
| openshift/hypershift | [#6183](https://github.com/openshift/hypershift/pull/6183), [#8839](https://github.com/openshift/hypershift/pull/8839) | Console capability in hosted clusters ([OTA-1956](https://redhat.atlassian.net/browse/OTA-1956)) |

### ⚠️ Key lesson from CMO and kube-apiserver-operator

Adding `capability.openshift.io/name: Console` to *existing* manifests during an upgrade causes CVO to **force-enable the Console capability** on clusters that had it disabled ([OCPBUGS-20331](https://issues.redhat.com/browse/OCPBUGS-20331)). Both CMO ([#2118](https://github.com/openshift/cluster-monitoring-operator/pull/2118) revert, [#2254](https://github.com/openshift/cluster-monitoring-operator/pull/2254) re-fix) and kube-apiserver-operator ([#1565](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1565)) had to rename their manifests to work around this. The runtime detection approach used in this PR avoids this pitfall entirely.

</details>

## Summary by CodeRabbit

- **New Features**
  - Added an OpenShift Console Quick Start for cert-manager, guiding users through creating an `Issuer` and `Certificate`, verifying the generated TLS Secret, and using certificates with Routes and Ingresses.
  - Added Console samples for ACME issuers, issuers, and certificates.
  - Console resources are automatically installed when the OpenShift Console capability is enabled.

- **Improvements**
  - Updated sample presentation and metadata for improved Console integration.


pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,144 @@
...
+			panic(fmt.Sprintf("console asset %s: %v", file, err))
+		}
+		c.consoleAssets = append(c.consoleAssets, consoleAsset{obj: obj, gvr: gvr})
+	}
+
+	return factory.New().
+		WithInformers(operatorClient.Informer()).
+		WithSync(c.sync).
+		ToController(consoleResourcesControllerName, c.eventRecorder)
+}
```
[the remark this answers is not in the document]
sebrandon1: Added `ResyncEvery(10*time.Minute)` to the controller factory builder so capability changes are detected even without an operator config event.

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,144 @@
...
+			ctx, c.dynamicClient, c.eventRecorder,
+			asset.obj, nil, asset.gvr, nil, nil,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to apply %s/%s: %w", asset.obj.GetKind(), asset.obj.GetName(), err)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
sebrandon1: Fixed — sync now continues applying all assets and returns aggregated errors via `errors.Join(errs...)`. Unit test `TestSyncAttemptsAllAssetsOnError` verifies all assets are attempted even when some fail.

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,152 @@
...
+	consoleAssetFiles = []string{
+		"console/cert-manager-acme-issuer-sample.yaml",
+		"console/cert-manager-certificate-sample.yaml",
+		"console/cert-manager-issuer-sample.yaml",
+		"console/cert-manager-example-quickstart.yaml",
+	}
+)
+
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consoleyamlsamples;consolequickstarts,verbs=get;list;watch;create;update;patch;delete
```
bharath-b-rh: Please remove unrequired permissions.
```suggestion
// +kubebuilder:rbac:groups=console.openshift.io,resources=consoleyamlsamples;consolequickstarts,verbs=get;create;update
```

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,152 @@
...
+		dynamicClient: dynamicClient,
+		eventRecorder: eventRecorder.WithComponentSuffix("console-resources"),
+	}
+
+	for _, file := range consoleAssetFiles {
+		data, err := assets.Asset(file)
+		if err != nil {
+			panic(fmt.Sprintf("failed to read console asset %s: %v", file, err))
+		}
+		obj := resourceread.ReadUnstructuredOrDie(data)
```
bharath-b-rh: I think it would be better to use typed clientset instead of dynamic client, which is available in `github.com/openshift/client-go/console/clientset/versioned` package.

config/console/cert-manager-acme-issuer-sample.yaml
```diff
@@ -2,8 +2,6 @@ apiVersion: console.openshift.io/v1
 kind: ConsoleYAMLSample
```
bharath-b-rh: I am bit confused, why the files in `config/console` are still being listed here, even after moved to `bindata/console`

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,152 @@
...
+		return false, fmt.Errorf("failed to get ClusterVersion: %w", err)
+	}
+
+	return slices.Contains(
+		version.Status.Capabilities.EnabledCapabilities,
+		configv1.ClusterVersionCapabilityConsole,
+	), nil
+}
+
+func gvrForKind(kind string) (schema.GroupVersionResource, error) {
```
bharath-b-rh: We shouldn't be needing this, if we split `consoleAssetFiles` for `ConsoleYAMLSampleGVR` and `ConsoleQuickStartGVR` and use respective typed client.

pkg/controller/certmanager/console_resources_test.go
```diff
@@ -0,0 +1,449 @@
...
+	got, err := c.hasConsoleCapability(context.Background())
+	if err != nil {
+		t.Fatalf("expected no error when ClusterVersion is not found (e.g. MicroShift), got: %v", err)
+	}
+	if got {
+		t.Error("expected false when ClusterVersion does not exist")
+	}
+}
+
+func TestHasConsoleCapabilityResourceNotFound(t *testing.T) {
```
bharath-b-rh: fake clientset from the client-go package can be used here.

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,152 @@
...
+
+	consoleAssetFiles = []string{
+		"console/cert-manager-acme-issuer-sample.yaml",
+		"console/cert-manager-certificate-sample.yaml",
+		"console/cert-manager-issuer-sample.yaml",
+		"console/cert-manager-example-quickstart.yaml",
+	}
+)
+
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
```
bharath-b-rh:
```suggestion
// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get
```

bindata/console/cert-manager-example-quickstart.yaml
```diff
@@ -3,7 +3,6 @@ kind: ConsoleQuickStart
 metadata:
   name: cert-manager-example
   annotations:
-    capability.openshift.io/name: Console
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
```
bharath-b-rh: nit: below annotations too are not required I think.
```suggestion
```

bharath-b-rh:
I was also thinking, why reysnc every 10m and check the CRD existence, when on certain cluster console CRDs will never be installed.
Please update the logic to use optional informers we have in-place for Infrastructure resource and the controller reconciles only when the CRD exists. And should on any cluster, the console operator be installed day-2(not sure if it's possible), one can restart the cert-manager operator to be pick up the change.

pkg/controller/certmanager/console_resources.go
sebrandon1: Done — narrowed to `verbs=get;create;update` for console resources.

pkg/controller/certmanager/console_resources.go
sebrandon1: Done — switched to the typed console clientset from `github.com/openshift/client-go/console/clientset/versioned`. The controller now uses `ConsoleYAMLSamples()` and `ConsoleQuickStarts()` typed clients with Get/Create/Update instead of `dynamic.Interface` + `ApplyUnstructuredResourceImproved`. Tests use `consolefake.NewClientset()`.

config/console/cert-manager-acme-issuer-sample.yaml
sebrandon1: These files are being *deleted* from `config/console/` — the diff shows them with `-` prefix because they're removed from the bundle path. The content has been moved to `bindata/console/` where it's embedded at build time and applied at runtime.

sebrandon1:
@bharath-b-rh Thanks for the thorough review! All comments addressed in the latest force-push:

**RBAC (lines 53-54):** Narrowed to `verbs=get` for ClusterVersions and `verbs=get;create;update` for console resources. Regenerated bundle via `make bundle`.

**Typed clientset (line 87):** Replaced `dynamic.Interface` + `ApplyUnstructuredResourceImproved` with the typed console clientset from `github.com/openshift/client-go/console/clientset/versioned`. The controller now uses `ConsoleYAMLSamples()` and `ConsoleQuickStarts()` typed clients directly.

**Asset files split (line 143):** Split `consoleAssetFiles` into `yamlSampleAssetFiles` and `quickStartAssetFiles`, decoded into typed `consolev1.ConsoleYAMLSample` and `consolev1.ConsoleQuickStart` structs. The `gvrForKind()` function and `consoleAsset` struct are no longer needed and have been removed.

**Tests (line 118):** Tests now use `consolefake.NewClientset()` from the typed console client-go package. Tests verify actual resource creation (List after sync shows correct counts) rather than just checking action counts.

**config/console files:** These appear in the diff because they are being deleted - the content was moved from `config/console/` (OLM bundle path) to `bindata/console/` (runtime-embedded). The deletion is the intended behavior.

**Quickstart annotations (line 8):** Removed the stale `include.release.openshift.io/*` annotations from the quickstart YAML.

sebrandon1:
@bharath-b-rh All review comments addressed in the latest force-push:

- **RBAC**: Narrowed to verbs=get for ClusterVersions and verbs=get;create;update for console resources
- **Typed clientset**: Replaced dynamic.Interface with typed console clientset from openshift/client-go/console
- **Asset split**: Separate yamlSampleAssetFiles and quickStartAssetFiles decoded into typed structs
- **Tests**: Using consolefake.NewClientset() with typed assertions
- **config/console**: Files are being deleted (moved to bindata/console for runtime embedding)
- **Annotations**: Removed stale include.release.openshift.io annotations from quickstart

sebrandon1:
All review comments addressed. Switched to typed console clientset, narrowed RBAC permissions, removed stale annotations, and split asset files by kind. Tests use consolefake.NewClientset().

pkg/controller/certmanager/console_resources.go
sebrandon1: Addressed in latest push.

pkg/controller/certmanager/console_resources_test.go
sebrandon1: Addressed in latest push.

pkg/controller/certmanager/console_resources.go
sebrandon1: Addressed in latest push.

bindata/console/cert-manager-example-quickstart.yaml
sebrandon1: Addressed in latest push.

sebrandon1:
Good call — updated to use startup-time CRD discovery instead of per-sync ClusterVersion polling.

**What changed:**
- At startup, `utils.NewResourceDiscoverer(ConsoleYAMLSampleGVR, ...)` checks if the Console CRDs exist
- Console controller is only created if CRDs are discovered; otherwise it is skipped entirely
- Removed `hasConsoleCapability` / ClusterVersion capability check and the `clusterversions` RBAC
- Day-2 console enablement is picked up on operator restart

This follows the same `ResourceDiscoverer` pattern already used for Infrastructure CRDs in `starter.go`. Cross-referenced [cluster-authentication-operator#589](https://github.com/openshift/cluster-authentication-operator/pull/589) which validates the startup-only approach (they use a TerminationController for day-2, but since console enablement is a one-time cluster decision, a restart is sufficient here).

test/e2e/console_resources_test.go
```diff
@@ -0,0 +1,135 @@
...
+	var (
+		ctx            context.Context
+		consoleEnabled bool
+	)
+
+	BeforeAll(func() {
+		ctx = context.Background()
+
+		By("detecting Console capability on the cluster")
+		enabled, err := hasConsoleCapability(ctx)
```
bharath-b-rh: `hasConsoleCapability()`  still depends on the ClusterVersions CR. Could we please use same as in the code which uses Discovery client.

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, desired := range c.quickStarts {
+		if err := c.applyQuickStart(ctx, desired); err != nil {
+			errs = append(errs, fmt.Errorf("failed to apply ConsoleQuickStart/%s: %w", desired.Name, err))
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (c *consoleResourcesController) applyYAMLSample(ctx context.Context, desired *consolev1.ConsoleYAMLSample) error {
```
bharath-b-rh: `applyYAMLSample()` and `applyQuickStart()` are almost same, except for the arg and the client used. I think we can common method for majority of code, or make use of generics.

pkg/controller/certmanager/console_resources.go
```diff
@@ -0,0 +1,160 @@
...
+) factory.Controller {
+	c := &consoleResourcesController{
+		consoleClient: consoleClient,
+		eventRecorder: eventRecorder.WithComponentSuffix("console-resources"),
+		yamlSamples:   loadAssets[consolev1.ConsoleYAMLSample](yamlSampleAssetFiles),
+		quickStarts:   loadAssets[consolev1.ConsoleQuickStart](quickStartAssetFiles),
+	}
+
+	return factory.New().
+		WithInformers(operatorClient.Informer()).
```
bharath-b-rh, line 90: We could use informers instead trigger reconcile for these every 10m, which is really not required I think. And this would require the list, watch permissions to be added back. I am more inclined towards watches.

@mytreya-rh Please let us know WDYT?

```suggestion
    WithInformers(
        operatorClient.Informer(),
        consoleInformerFactory.Console().V1().ConsoleYAMLSamples().Informer(),
        consoleInformerFactory.Console().V1().ConsoleQuickStarts().Informer(),
    ).
    WithSync(c.sync).
    ToController(consoleResourcesControllerName, c.eventRecorder)
```

bharath-b-rh (commented):
@sebrandon1 Thank you for accepting the suggestions! I just had few more. PTAL.

Also I am reviewing https://github.com/openshift/release/pull/81900, and I think we could test changes here with the above job.

test/e2e/console_resources_test.go
sebrandon1: I'll take a look.

pkg/controller/certmanager/console_resources.go
sebrandon1: Thanks for the suggestion. I looked into this and found precedent for both approaches within [openshift/console-operator](https://github.com/openshift/console-operator) (the owner of these CRDs):

**Option A: Keep `ResyncEvery` (current)**
- console-operator's [UpgradeNotification controller](https://github.com/openshift/console-operator/blob/main/pkg/console/controllers/upgradenotification/controller.go#L67) uses this exact pattern — creates `ConsoleNotification` resources with `ResyncEvery(1m)` + config informers, without watching its own resources.
- Simpler: no additional vendoring needed (`console/informers` and `console/listers` are not currently vendored in cert-manager-operator).
- Current RBAC (`get;create;update`) is sufficient.

**Option B: Add console informers (your suggestion)**
- console-operator's [CLIDownloads controller](https://github.com/openshift/console-operator/blob/main/pkg/console/controllers/clidownloads/controller.go#L91-L93) uses `WithInformers(consoleCLIDownloadsInformers.Informer())` + `ResyncEvery(1m)`.
- More reactive — detects external changes immediately rather than up to 10 minutes later.
- Requires: vendoring `console/informers` + `console/listers`, adding `list;watch` RBAC, wiring a conditional `consoleinformers.SharedInformerFactory` through `starter.go` (using our existing [`OptionalInformer` pattern](https://github.com/openshift/cert-manager-operator/blob/master/pkg/operator/utils/apidiscovery.go)).
- Note: [OCPBUGS-95609](https://issues.redhat.com/browse/OCPBUGS-95609) documented console-operator excessive rollouts from informer-based reconciliation — events need proper filtering.

Since these are purely operator-managed resources (no admin is expected to modify them), I lean toward Option A for simplicity. Happy to go either way — waiting on @mytreya-rh's input as well.

test/e2e/console_resources_test.go
sebrandon1: Done.

pkg/controller/certmanager/console_resources.go
sebrandon1: Done.

pkg/controller/certmanager/console_resources.go
mytreya-rh: i would ideally prefer both a watch and periodic resync for both responsiveness and eventual consistency :-)
But in this case, Bharath and i re-discussed and since these CRs are not something users maybe aware of, i feel it is more important to be eventually consistent,  i.e. Option A

bharath-b-rh:
cc: @snarayan-redhat for the docs ack.

```
## Fixed issues

Before this update, the cert-manager Operator for Red Hat OpenShift installation failed on clusters with the Console capability disabled because the OLM bundle included ConsoleYAMLSample and ConsoleQuickStart resources that require the console.openshift.io APIs. With this release, the Operator creates these Console resources at runtime only when those APIs are available, so installation succeeds on Console-less clusters. (OCPBUGS-85579)
```

pkg/controller/certmanager/console_resources.go
sebrandon1: Thanks @mytreya-rh — Option A it is. We'll keep `ResyncEvery(10*time.Minute)` as the reconciliation mechanism, no console informer watches needed. Agreed that eventual consistency is the right tradeoff for operator-managed resources that users don't interact with directly.

bharath-b-rh:
/lgtm
/approve

adding below labels based on the CI results
/label px-approved
/label qe-approved

PR #466, opened 2026-08-03
CM-1225: Apply cluster TLS profile to trust-manager, operator metrics, and operand HTTPS metrics

arun717:
## Summary
- Apply cluster TLS security profile to trust-manager webhook serving TLS
- Honor cluster TLS profile on the operator metrics server (`:8443`)
- Enable HTTPS metrics TLS on cert-manager controller/webhook/cainjector operands
- Claim/sync CSV `tls-profiles` feature annotation

## Test plan
- [x] `/test tls-scanner`
- [x] `/test e2e-operator-tech-preview` (TLS profile e2e)
- [x] `/test unit`
- [x] `/test verify`

## Summary by CodeRabbit

* **New Features**
  * Added cluster-wide TLS profile support for cert-manager, trust-manager, and operator-serving endpoints.
  * Enabled HTTPS metrics for cert-manager controller, webhook, and cainjector components.
  * Added automatic serving certificate integration for OpenShift deployments.
  * Added trust-manager TLS configuration based on the active cluster profile.

* **Bug Fixes**
  * Improved TLS 1.3 handling by avoiding incompatible cipher-suite settings.
  * Added automatic updates when cluster TLS configuration changes.
  * Improved metrics security through dynamic serving certificates and permissions.

### DIFF
--- a/bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
+++ b/bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
@@ -21,6 +21,7 @@ spec:
       annotations:
         prometheus.io/path: /metrics
         prometheus.io/port: "9402"
+        prometheus.io/scheme: https
         prometheus.io/scrape: "true"
       labels:
         app: cainjector
--- a/bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
+++ b/bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
@@ -21,7 +21,6 @@ spec:
       annotations:
         prometheus.io/path: /metrics
         prometheus.io/port: "9402"
-        prometheus.io/scheme: https
         prometheus.io/scrape: "true"
       labels:
         app: cainjector
--- a/bindata/cert-manager-deployment/controller/cert-manager-metrics-dynamic-serving-rb.yaml
+++ b/bindata/cert-manager-deployment/controller/cert-manager-metrics-dynamic-serving-rb.yaml
@@ -1,0 +1,25 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  labels:
+    app: cert-manager
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: cert-manager
+    app.kubernetes.io/name: cert-manager
+    app.kubernetes.io/version: v1.20.3
+  name: cert-manager-metrics-dynamic-serving
+  namespace: cert-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: cert-manager-metrics-dynamic-serving
+subjects:
+  - kind: ServiceAccount
+    name: cert-manager
+    namespace: cert-manager
+  - kind: ServiceAccount
+    name: cert-manager-webhook
+    namespace: cert-manager
+  - kind: ServiceAccount
+    name: cert-manager-cainjector
+    namespace: cert-manager
--- a/bundle/manifests/cert-manager-operator-controller-manager-metrics-service_v1_service.yaml
+++ b/bundle/manifests/cert-manager-operator-controller-manager-metrics-service_v1_service.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: cert-manager-operator-serving-cert
   creationTimestamp: null
   labels:
     app.kubernetes.io/created-by: cert-manager-operator
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -141,8 +141,79 @@ local path(item) =
   // else, leave it at the top-level
   else 'bindata/cert-manager-deployment/' + item.metadata.name + '-' + suffix[item.kind] + '.yaml';
 
+// OpenShift extra: not in the default upstream install YAML. Metrics HTTPS is
+// enabled by the operator via --metrics-dynamic-serving-* against a shared
+// cert-manager-metrics-ca Secret. Re-emit these on every
+// hack/update-cert-manager-manifests.sh run so they are not lost when
+// bindata/cert-manager-deployment is regenerated.
+local metricsDynamicServingCASecretName = 'cert-manager-metrics-ca';
+local metricsDynamicServingName = 'cert-manager-metrics-dynamic-serving';
+
+local operandVersion(manifest) = [
+  item.metadata.labels['app.kubernetes.io/version']
+  for item in manifest
+  if std.type(item) == 'object'
+     && 'metadata' in item
+     && 'labels' in item.metadata
+     && 'app.kubernetes.io/version' in item.metadata.labels
+][0];
+
+local controllerLabels(version) = {
+  app: 'cert-manager',
+  'app.kubernetes.io/component': 'controller',
+  'app.kubernetes.io/instance': 'cert-manager',
+  'app.kubernetes.io/name': 'cert-manager',
+  'app.kubernetes.io/version': version,
+};
+
+local extraManifests(manifest) =
+  local version = operandVersion(manifest);
+  [
+    {
+      apiVersion: 'rbac.authorization.k8s.io/v1',
+      kind: 'Role',
+      metadata: {
+        labels: controllerLabels(version),
+        name: metricsDynamicServingName,
+        namespace: sourceOperandNamespace,
+      },
+      rules: [
+        {
+          apiGroups: [''],
+          resourceNames: [metricsDynamicServingCASecretName],
+          resources: ['secrets'],
+          verbs: ['get', 'list', 'watch', 'update'],
+        },
+        {
+          apiGroups: [''],
+          resources: ['secrets'],
+          verbs: ['create'],
+        },
+      ],
+    },
+    {
+      apiVersion: 'rbac.authorization.k8s.io/v1',
+      kind: 'RoleBinding',
+      metadata: {
+        labels: controllerLabels(version),
+        name: metricsDynamicServingName,
+        namespace: sourceOperandNamespace,
+      },
+      roleRef: {
+        apiGroup: 'rbac.authorization.k8s.io',
+        kind: 'Role',
+        name: metricsDynamicServingName,
+      },
+      subjects: [
+        { kind: 'ServiceAccount', name: 'cert-manager', namespace: sourceOperandNamespace },
+        { kind: 'ServiceAccount', name: 'cert-manager-webhook', namespace: sourceOperandNamespace },
+        { kind: 'ServiceAccount', name: 'cert-manager-cainjector', namespace: sourceOperandNamespace },
+      ],
+    },
+  ];
+
 // top level function (aka 'main')
 function(manifest) {
   [std.strReplace(path(item), ':', '-')]: processManifests(cleanupHelmLabels(item))
-  for item in manifest
+  for item in (manifest + extraManifests(manifest))
 }
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -2,23 +2,86 @@ package operator
 
 import (
 	"context"
+	"math/rand"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apiserver/pkg/server"
+	"k8s.io/component-base/logs"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
 
 	"github.com/openshift/cert-manager-operator/pkg/operator"
+	"github.com/openshift/cert-manager-operator/pkg/tlsprofile"
 	"github.com/openshift/cert-manager-operator/pkg/version"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
-	"github.com/spf13/cobra"
-	"k8s.io/utils/clock"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/serviceability"
 )
 
 func NewOperator() *cobra.Command {
-	cmd := controllercmd.NewControllerCommandConfig(
+	cc := controllercmd.NewControllerCommandConfig(
 		"cert-manager-operator",
 		version.Get(),
 		operator.RunOperator,
 		clock.RealClock{},
-	).NewCommandWithContext(context.TODO())
+	)
+
+	cmd := cc.NewCommandWithContext(context.TODO())
 	cmd.Use = "start"
 	cmd.Short = "Start the cert-manager Operator"
+
+	// Replace the default Run so we can apply the cluster TLS profile to the
+	// metrics serving config before the HTTPS listener is created.
+	cmd.Run = func(cmd *cobra.Command, args []string) {
+		rand.Seed(time.Now().UTC().UnixNano())
+		logs.InitLogs()
+		defer logs.FlushLogs()
+		defer serviceability.BehaviorOnPanic(os.Getenv("OPENSHIFT_ON_PANIC"), version.Get())()
+		defer serviceability.Profile(os.Getenv("OPENSHIFT_PROFILE")).Stop()
+		serviceability.StartProfiler()
+
+		shutdownCtx, cancel := context.WithCancel(context.Background())
+		shutdownHandler := server.SetupSignalHandler()
+		go func() {
+			defer cancel()
+			<-shutdownHandler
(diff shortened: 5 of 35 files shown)


arun717:
Test Results on a live cluster:

Cluster TLS config is writable (TLSAdherence + APIServer profile)
Profile/adherence scenarios: 
- baseline → no args
- StrictAllComponents + Intermediate/Modern/Custom → args applied
- LegacyAdheringComponentsOnly → args removed

tls-scanner: 0 compliance failures for trust-manager :6443, operator :8443, and operand HTTPS metrics

In scope of what we tested:
----------------------------------------
- Trust-manager webhook TLS
- Operator metrics TLS
- cert-manager / webhook / cainjector HTTPS metrics TLS
- Honoring tlsAdherence (Strict vs Legacy)

Not covered (by design or not in this branch):
------------------------------------------------------------------
- Trust-manager :9402 stays plain HTTP (NO_TLS)
- TLS Groups / curve preferences
- istio-csr

arun717:
Left-out e2e (by design): 
live operator :8443 TLS handshake, in-cluster metrics-CA RBAC/secret readiness, and serving-cert Service/volume probes.

Why: 
- trust-manager and cert-manager operand TLS-profile behavior is validated via Deployment args/annotations and Day-2 reconcile (including Legacy vs always-on HTTPS metrics).
- Operator metrics :8443 TLS is applied to HTTPServingInfo only at process start, so a meaningful e2e would need an operator restart, metrics dial, and serving-CA trust - high cost/flake versus unit coverage of ApplyToHTTPServingInfo / ApplyClusterProfileToHTTPServingInfo and ResolveHonoredTLSProfile adherence. 
- Metrics-CA Role/RoleBinding and serving-cert Service/volume wiring are static manifest concerns better suited to UT/Manual QE than cluster waits.

Accepted residual risk: no live proof that the running metrics listener negotiates Modern/Intermediate after restart; no e2e proof of metrics-CA secret minting.

bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
```diff
@@ -21,6 +21,7 @@ spec:
       annotations:
         prometheus.io/path: /metrics
         prometheus.io/port: "9402"
+        prometheus.io/scheme: https
```
bharath-b-rh: Was this manually added? If so this needs to be handled through `hack/update-cert-manager-manifests.sh` to patch the content.

Also, IIRC, port 9402 is for http APIs, was the port updated to handle https APIs?

bindata/cert-manager-deployment/controller/cert-manager-metrics-dynamic-serving-rb.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
bharath-b-rh, line 1: Is the manifest added manually? If so the manifest update script must be updated to preserve this. Is this manifest not already available in upstream chart?

bundle/manifests/cert-manager-operator-controller-manager-metrics-service_v1_service.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
```
bharath-b-rh, line 4: This is going to be problem when and if we are going to use the cert-manager for internal certificates too. We will need to provide an option for users to configure the TLS keys for operator metrics too, as an arg, probably reference to secret.

pkg/cmd/operator/cmd.go
```diff
@@ -2,23 +2,89 @@ package operator
...
-	"github.com/spf13/cobra"
-	"k8s.io/utils/clock"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/serviceability"
 )
 
 func NewOperator() *cobra.Command {
-	cmd := controllercmd.NewControllerCommandConfig(
+	cc := controllercmd.NewControllerCommandConfig(
```
bharath-b-rh, line 25: I understand the need for changes in this file, specifically duplicating the whole `Run` just to replace `StartController` with custom `startControllerWithClusterTLS` for adding TLS specifics.

But I think this can be avoided and will help other operators as well, if the library-go can be updated with a hook function to patch the TLS configs.

Something like
```
func (c *ControllerCommandConfig) WithServingInfoHook(
    fn func(ctx context.Context, restConfig *rest.Config, serving *configv1.HTTPServingInfo) error,
) *ControllerCommandConfig
```

pkg/controller/certmanager/deployment_metrics_tls.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+
+	container := &deployment.Spec.Template.Spec.Containers[0]
+	container.Args = common.MergeContainerArgs(container.Args, extra)
+
+	if deployment.Spec.Template.Annotations == nil {
+		deployment.Spec.Template.Annotations = map[string]string{}
+	}
+	deployment.Spec.Template.Annotations["prometheus.io/scheme"] = "https"
```
bharath-b-rh, line 37: We are adding it here and in the bindata as well. I think manual edit to bindata can be avoided with this.

pkg/controller/certmanager/deployment_metrics_tls_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"strings"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestWithOperandMetricsTLS(t *testing.T) {
```
bharath-b-rh, line 15: Could we add some test cases for below

- error: no containers
- no-op: unknown deployment name
- nil annotations → scheme set, map created
- preserve prometheus.io/port
- preserve --v=2
- idempotent second apply
- override stale --metrics-dynamic-serving-dns-names
- symmetric wantArgs for webhook/cainjector (add namespace assertion)

pkg/controller/trustmanager/controller.go
```diff
@@ -130,6 +134,11 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			object.GetName() == common.TrustedCABundleConfigMapName
 	})
 
+	// Reconcile when the cluster APIServer TLS profile or adherence changes.
```
bharath-b-rh: If we want to reconcile only on the changes to TLSAdherence and TLSSecurityProfile, we could add a predicate helper like below to use with `predicate.Funcs`, instead of basic check.

```
func ClusterAPIServerTLSConfigChanged(old, new *configv1.APIServer) bool {
	if old == nil || new == nil {
		return old != new
	}
	return old.Spec.TLSAdherence != new.Spec.TLSAdherence ||
		!equality.Semantic.DeepEqual(old.Spec.TLSSecurityProfile, new.Spec.TLSSecurityProfile)
}
```

pkg/controller/trustmanager/deployment_tls_test.go
```diff
@@ -0,0 +1,340 @@
...
+		{
+			name: "intermediate sets min version and ciphers",
+			spec: &configv1.TLSProfileSpec{
+				Ciphers:       []string{"ECDHE-RSA-AES128-GCM-SHA256"},
+				MinTLSVersion: configv1.VersionTLS12,
+			},
+			wantKeys: []string{"--tls-min-version", "--tls-cipher-suites"},
+		},
+		{
+			name: "modern tls13 omits cipher suites",
```
bharath-b-rh, line 53: Could we add some test cases for different profiles, especially for custom type?

pkg/tlsprofile/tlsprofile.go
```diff
@@ -90,7 +96,62 @@ func CertManagerOperandMetricsTLSArgs(spec *configv1.TLSProfileSpec) []string {
...
+		if len(serving.CipherSuites) == 0 {
+			// Keep a non-empty list so SetRecommendedHTTPServingInfoDefaults does
+			// not overwrite with Intermediate defaults; TLS 1.3 ignores these.
+			serving.CipherSuites = []string{"TLS_AES_128_GCM_SHA256"}
+		}
+		return nil
+	}
+	iana := libgocrypto.OpenSSLToIANACipherSuites(spec.Ciphers)
+	if len(spec.Ciphers) > 0 && len(iana) == 0 {
+		return fmt.Errorf("no cipher suites after OpenSSL→IANA mapping")
```
bharath-b-rh, line 153: Should this be an error, is it being really strict? We could just log I guess.

pkg/controller/certmanager/deployment_metrics_tls.go
```diff
@@ -0,0 +1,59 @@
...
+	if len(deployment.Spec.Template.Spec.Containers) == 0 {
+		return fmt.Errorf("deployment %s/%s has no containers", deployment.Namespace, deployment.Name)
+	}
+
+	extra, ok := operandMetricsTLSArgs(deployment.Name)
+	if !ok {
+		return nil
+	}
+
+	container := &deployment.Spec.Template.Spec.Containers[0]
```
bharath-b-rh, line 31: nit: not an issue today. But should we be hardcoding the index here.

bindata/cert-manager-deployment/cainjector/cert-manager-cainjector-deployment.yaml
arun717: Thanks for pointing it out. Yes, that annotation was added by hand in bindata. It shouldn’t live there — hack/update-cert-manager-manifests.sh would overwrite it, and jsonnet doesn’t patch it.

The operator already sets prometheus.io/scheme: https in **withOperandMetricsTLS** when it injects --metrics-dynamic-serving-*. I’ve dropped it from all three operand deployments (cainjector, controller, webhook) and regenerated pkg/operator/assets/bindata.go. Prometheus will still scrape HTTPS on that annotation once the TLS flags are injected at reconcile time.

Port 9402 does not need to change. That is still the metrics listener; TLS is enabled on it via the dynamic serving CA flags. Scrapers use scheme=https on the same port.

bindata/cert-manager-deployment/controller/cert-manager-metrics-dynamic-serving-rb.yaml
arun717: Yes, this is operator-owned extra, not in the default upstream install YAML. I’ve added the Role/RoleBinding to jsonnet/main.jsonnet so hack/update-cert-manager-manifests.sh re-emits them on each bump.

pkg/cmd/operator/cmd.go
arun717: Agreed. Copying Run / StartController is only because library-go has no way to patch HTTPServingInfo before the metrics listener is created.

A WithServingInfoHook on ControllerCommandConfig (invoked in StartController before WithServer) is the right fix and would help other operators. I’ll keep the local copy in this PR so we can land TLS profile support, and follow up with a library-go change plus a vendor bump to drop the duplication.

pkg/controller/certmanager/deployment_metrics_tls.go
arun717: Agreed that this hook is the source of truth. The bindata copy was redundant and would be overwritten by hack/update-cert-manager-manifests.sh.

I dropped prometheus.io/scheme: https from the cainjector, controller, and webhook bindata deployments (and regenerated pkg/operator/assets/bindata.go). Prometheus will still scrape HTTPS once withOperandMetricsTLS injects the annotation with --metrics-dynamic-serving-*.

pkg/controller/certmanager/deployment_metrics_tls_test.go
arun717: done.

pkg/controller/trustmanager/controller.go
arun717: Good catch. using predicate.Funcs and making the change.

pkg/controller/trustmanager/deployment_tls_test.go
arun717: Sure. Adding the test cases.

pkg/tlsprofile/tlsprofile.go
arun717: It is strict on purpose. This only happens when a Custom profile’s cipher list maps to nothing. Logging and continuing would leave CipherSuites empty, then library-go would fill Intermediate defaults and we would serve a different policy than configured.

Returning an error fails loud: the operator does not start, and the admin can fix the Custom list. That matches the requirement of “do not proceed with an unknown TLS configuration.”

FYI: Named profiles (Old, Intermediate, and Modern) are unaffected. Custom profile is the one where an admin types cipher names by hand.
Also, partial lists are already lenient. If some names map, those are kept and there is no error.

pkg/controller/certmanager/deployment_metrics_tls.go
arun717: These operands are single-container, and the other cert-manager deployment hooks already use index 0. Let me add a require len == 1 before using Containers[0] to tighten this for now.

Name lookup is the better long-term pattern if we add a sidecar, but then we would have to change all operand hooks together.

jsonnet/main.jsonnet
```diff
@@ -141,8 +141,79 @@ local path(item) =
   // else, leave it at the top-level
   else 'bindata/cert-manager-deployment/' + item.metadata.name + '-' + suffix[item.kind] + '.yaml';
 
+// OpenShift extra: not in the default upstream install YAML. Metrics HTTPS is
```
bharath-b-rh, line 144: I don't we are creating any custom static manifests for cert-manager. Maybe these manifests are not present in upstream in current release or there should be some reason why these ended up in bindata. Please check once.

pkg/controller/trustmanager/controller.go
```diff
@@ -59,6 +62,7 @@ type Reconciler struct {
 // +kubebuilder:rbac:groups=trust.cert-manager.io,resources=bundles,verbs=get;list;watch
 // +kubebuilder:rbac:groups=trust.cert-manager.io,resources=bundles/finalizers,verbs=update
 // +kubebuilder:rbac:groups=trust.cert-manager.io,resources=bundles/status,verbs=patch
+// +kubebuilder:rbac:groups=config.openshift.io,resources=apiservers,verbs=get;list;watch
```
bharath-b-rh: This is already added for the cert-manager operand correct? So the operator would have this permission already, not required to add here again.

jsonnet/main.jsonnet
arun717: The reason for adding it is that **hack/update-cert-manager-manifests.sh** starts from upstream cert-manager.yaml, then jsonnet rewrites those objects and adds two more:

- cert-manager-metrics-dynamic-serving Role
- matching RoleBinding (controller, webhook, cainjector SAs)

Those two are not in the default upstream install YAML. 
They exist because the operator turns on HTTPS metrics with --metrics-dynamic-serving-* against cert-manager-metrics-ca (withOperandMetricsTLS). 

Dynamic serving needs RBAC to create/update that Secret. But the regenerate script deletes bindata/cert-manager-deployment and writes only what jsonnet emits. So, without extraManifests(), those YAML files disappear on the next bump.

That is why they are synthesized in jsonnet instead of being hand-edited in bindata.

**TL;DR**
These two RBAC objects are operator-added (metrics HTTPS), not from upstream, so jsonnet has to emit them or they vanish on regenerate.

pkg/controller/trustmanager/controller.go
arun717: yes. Thanks for catching this. Deleting line from this file.

jsonnet/main.jsonnet
```diff
@@ -141,8 +141,79 @@ local path(item) =
...
+
+local controllerLabels(version) = {
+  app: 'cert-manager',
+  'app.kubernetes.io/component': 'controller',
+  'app.kubernetes.io/instance': 'cert-manager',
+  'app.kubernetes.io/name': 'cert-manager',
+  'app.kubernetes.io/version': version,
+};
+
+local extraManifests(manifest) =
```
bharath-b-rh: nit: can be done in follow-up PR as well.

I am not comfortable having the manifests here in this code, this could cause confusion and while debugging in the future is an unlikely place to be checked for the manifest ToS.
Could we add another dir under `bindata/cert-manager-deployment` for these two manifests and selectively delete dirs something like [here](https://github.com/bharath-b-rh/external-secrets-operator/blob/8502103c857bf59d8f208ea0ad0560817a3eb6d4/hack/update-external-secrets-manifests.sh#L51-L58)

bharath-b-rh (approved):
/lgtm
/approve

jsonnet/main.jsonnet
arun717: Handling the changes in this PR.

bharath-b-rh (approved):
/lgtm

PR #477, opened 2026-08-19
ai-eval: DNM, CM-735: Fix intermittent unreconciled IstioCSR CR due to dual-cache race

swghosh:
## What

Fixes the intermittent failure where a newly created `IstioCSR` CR is not reconciled: its `.status` stays empty and the `cert-manager-istio-csr` Deployment is never created, with the operator logging:

```
"istiocsr.openshift.operator.io object not found, skipping reconciliation"
```

Restarting the operator pod resolves it, which pointed at a cache-consistency issue.

## Root cause

The controller used **two independent informers for the same `IstioCSR` resource**:

1. The manager's default cache drove reconcile events via `For(&IstioCSR{})`.
2. A separate label-filtered custom cache (built in `BuildCustomClient`) backed the reconciler's reads (`r.Get`).

On a freshly created `IstioCSR`, the manager cache could enqueue a reconcile before the custom cache observed the object. `Reconcile()` then read from the custom cache, got `NotFound`, and returned **without a requeue**. Since the object already existed, no further event reached the custom cache, so the CR stayed stuck until the pod restarted (forcing a LIST-based resync).

## Fix

Source the primary `IstioCSR` watch from the **same custom cache** that backs `r.Get`, using `WatchesRawSource(source.Kind(r.cache, ...))` instead of `For()`. The informer that triggers a reconcile is now the same one the read is served from, so the object is guaranteed present when its event fires. This also removes the redundant second `IstioCSR` informer on the manager's default cache.

- `BuildCustomClient` / `NewClient` now return the created `cache.Cache`.
- `Reconciler` gains a `cache` field, populated in `New`.
- `SetupWithManager` uses `WatchesRawSource` with a typed enqueue handler and generation-changed predicate; `.Named(ControllerName)` (already present) is required once `For()` is dropped.

## Testing

- `go build ./...` — passes
- `go vet ./pkg/controller/istiocsr/...` — clean
- `go test ./pkg/controller/istiocsr/...` — passes

Jira: https://issues.redhat.com/browse/CM-735

---
Always review AI generated responses prior to use.
Generated with [Claude Code](https://claude.com/claude-code) via openshift-developer plugin

## Summary by CodeRabbit

* **Improvements**
  * Improved controller resource monitoring with more targeted event filtering.
  * Reduced unnecessary processing during reconciliation for better operational efficiency.
  * Improved handling of controller cache initialization and resource access errors.
  * Reconciliation now responds more consistently to relevant resource updates.

### DIFF
--- a/pkg/controller/istiocsr/client.go
+++ b/pkg/controller/istiocsr/client.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/util/retry"
 
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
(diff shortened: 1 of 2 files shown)


swghosh:
/close

dummy PR created by [/jira:solve](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/jira#jirasolve---analyze-and-solve-jira-issues) for evals.

PR #478, opened 2026-08-19
CM-763: Fix infinite update loop for user-defined network policies

swghosh:
## Summary

- Fix infinite reconciliation loop in `CertManagerNetworkPolicyUserDefinedController` that caused continuous `NetworkPolicyUpdated` events every 400-800ms when multiple user-defined network policies were configured
- Add `apiequality.Semantic.DeepEqual` comparison of existing vs desired spec and labels before issuing an update — skip the write and event when nothing changed
- Make the `UserNetworkPolicyReconcileSuccess` event conditional on actual changes to eliminate steady-state event noise

## Root Cause

`createOrUpdateNetworkPolicy` unconditionally called `Update()` on every sync cycle for existing policies without comparing whether the spec or labels actually differed. Each update triggered informer re-queues, creating a tight reconciliation loop.

## Test plan

- [x] Unit tests for `createOrUpdateNetworkPolicy` covering create, no-op update, and spec-change update scenarios
- [ ] Manual verification: enable `defaultNetworkPolicy: "true"` with multiple `networkPolicies[]`, confirm no repeated `NetworkPolicyUpdated` events after initial creation
- [ ] Verify `NetworkPolicyCreated` events still fire on first policy creation
- [ ] Verify `NetworkPolicyUpdated` events fire when spec is actually changed

Always review AI generated responses prior to use.
Generated with [Claude Code](https://claude.com/claude-code) via openshift-developer plugin

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary updates when network policies are already up to date.
  * Success events are now emitted only when a policy is created or changed.
  * Improved error reporting for network policy operations.

* **Tests**
  * Added coverage for policy creation, unchanged policies, and updates after specification changes.

### DIFF
--- a/pkg/controller/deployment/cert_manager_networkpolicy.go
+++ b/pkg/controller/deployment/cert_manager_networkpolicy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


swghosh:
/close

dummy PR created by [/jira:solve](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/jira#jirasolve---analyze-and-solve-jira-issues) for evals.

PR #494, opened 2026-09-10
CM-1367: Add TrustManager filterNonCACerts API

arun717:
## Summary

- Add `spec.trustManagerConfig.filterNonCACerts` (`Enabled`/`Disabled`, default `Disabled`). When Enabled, the operator sets `--filter-non-ca-certs=true` the same way as `filterExpiredCertificates`.
- Add `spec.trustManagerConfig.webhookTLS`:
  - `certificateDuration` — optional webhook TLS Certificate duration (Helm `app.webhook.tls.certificate.duration`). When unset, cert-manager's default duration is used.
  - `approverPolicy.policy` — `Enabled`/`Disabled` (default `Disabled`). When Enabled, the operator creates `CertificateRequestPolicy` `trust-manager-policy` plus ClusterRole/ClusterRoleBinding so the cert-manager ServiceAccount can `use` that policy (Helm `app.webhook.tls.approverPolicy.enabled`). Nothing is created unless Enabled. If Enabled without the CRP CRD installed, reconcile fails until approver-policy is installed or policy is set to Disabled.
- Enabled/Disabled TrustManager fields are typed as shared `Mode` (`v1alpha1.Enabled` / `v1alpha1.Disabled`): `filterExpiredCertificates`, `filterNonCACerts`, `defaultCAPackage.policy`, and `webhookTLS.approverPolicy.policy`. `SecretTargetsPolicy` is unchanged (`Disabled`/`Custom`).
- `TrustManagerStatus` no longer copies spec (trust namespace, secret targets, default CA, filter policies). Status keeps conditions and observed `trustManagerImage`.
- No operand version bump in this PR.

**Depends on https://github.com/openshift/cert-manager-operator/pull/495** (trust-manager **v0.25.0**). `--filter-non-ca-certs` exists only from **v0.21.0**; merging this onto v0.20.3 and setting Enabled will CrashLoop the operand. Merge 495 first.

Jira: https://issues.redhat.com/browse/CM-1367

## Test plan

- [x] `go test ./pkg/controller/trustmanager/... ./api/operator/v1alpha1/...`
- [x] CRD testsuite cases for `filterNonCACerts`
- [x] Unit tests for deployment args
- [x] Unit tests for webhook certificate duration reconcile (set / skip defaulted)
- [x] Unit tests for approver-policy CRP + RBAC (disabled skip, apply, missing CRD)
- [x] E2e: args + Bundle Group 7 (CA + leaf, filter on/off) — CI needs 495’s operand
[...]

PR #495, opened 2026-09-10
CM-1367: Bump trust-manager operand to v0.25.0

arun717:
## Summary

- Bump the trust-manager operand from **v0.20.3** to **v0.25.0** (Makefile, `RELATED_IMAGE` / CSV, Helm bindata, Bundle CRD).
- Apply Helm `trust-manager-cluster-view` (read-only `bundles` + `aggregate-to-cluster-reader`). This stays a **separate** ClusterRole so cluster-reader does not inherit operand write rules.
- Align webhook CA injection with Helm 0.25: `cert-manager.io/inject-ca-from-secret` on `trust-manager-tls`, plus Certificate `secretTemplate` `cert-manager.io/allow-direct-injection: "true"`.
- Bump operator/test/tools modules to **Kubernetes v0.36.4** and **openshift/api `release-5.0`** (`31af9f93e31e`), with matching `openshift/client-go`, `library-go`, and controller-runtime **v0.24.1**.
- Move `test/go.mod` to **trust-manager v0.25.0** and update e2e Bundle builders for the v0.25 API (`InLine` string, value `TargetMetadata`). Upstream trust-manager still requires k8s 0.37; the test module replace-pins k8s 0.36.4 for OpenShift 5.0.

This is the operand half of [CM-1367](https://redhat.atlassian.net/browse/CM-1367). The TrustManager `filterNonCACerts` API lives in **https://github.com/openshift/cert-manager-operator/pull/494**. Merge **this PR first**: `--filter-non-ca-certs` exists only from v0.21.0, so 494’s Enabled path will CrashLoop a v0.20.3 operand.

Jira: https://issues.redhat.com/browse/CM-1367

## Notes for reviewers

- **Bundle CRD diff** is regenerated with `make bundle`. Content is the upstream Bundle schema (CEL, length limits, descriptions), not a YAML reformat. Helm output stays in `config/crd/bases/`; do not copy it onto `bundle/manifests/`.
- **`test/go.mod` uses `github.com/cert-manager/trust-manager v0.25.0`.** That module requires k8s.io/* v0.37.0; replace directives keep the selected Kubernetes/client stack at **v0.36.4** to match the operator and OpenShift 5.0.
- **`images/ci/trustmanager.Dockerfile`** sets `RELEASE_BRANCH=v0.25.0` and clones `openshift/cert-manager-trust-manager`. That fork’s latest tag is still **v0.20.3**; the Dockerfile clone will fail until midstream has the tag. CSV/`RELATED_IMAGE` uses `quay.io/jetstack/trust-manager:v0.25.0`.
- **cluster-view** is applied at the end of RBAC reconcile. `UpdateResourceLabels` replaces labels, so the aggregate label is set again after that.

## Test plan

- [x] Confirm CSV / manager related image is `quay.io/jetstack/trust-manager:v0.25.0`
- [ ] TrustManager Ready on v0.25.0 with default args (no `--filter-non-ca-certs`)
- [ ] Webhook annotation is `cert-manager.io/inject-ca-from-secret=cert-manager/trust-manager-tls`
- [ ] `oc get clusterrole trust-manager-cluster-view` and cluster-reader aggregation
- [x] Operator `go.mod` pins `k8s.io/* v0.36.4` and `github.com/openshift/api` release-5.0
- [x] E2e compiles against trust-manager v0.25.0 Bundle types (`go test -tags e2e -count=0 ./test/...`)
- [ ] CI e2e that already covers TrustManager (no new operator API in this PR)

## Summary by CodeRabbit

- **New Features**
  - Upgraded trust-manager components to version 0.25.0.
  - Added a cluster-view permission role, enabling read access to Bundle resources.
  - Added direct injection support for generated certificate secrets.
  - Added automatic private-key rotation behavior for the trust-manager certificate.
  - Improved webhook CA injection using the trust-manager TLS secret.

- **Validation**
  - Strengthened Bundle configuration validation, including source, target, format, length, and metadata constraints.

### DIFF
--- a/bindata/trust-manager/resources/certificate_trust-manager.yml
+++ b/bindata/trust-manager/resources/certificate_trust-manager.yml
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: cert-manager-trust-manager
     app.kubernetes.io/instance: cert-manager-trust-manager
-    app.kubernetes.io/version: "v0.20.3"
+    app.kubernetes.io/version: "v0.24.0"
     app.kubernetes.io/managed-by: cert-manager-operator
     app.kubernetes.io/part-of: cert-manager-operator
 spec:
@@ -15,6 +15,8 @@ spec:
   dnsNames:
     - "trust-manager.cert-manager.svc"
   secretName: trust-manager-tls
+  privateKey:
+    rotationPolicy: Always
   revisionHistoryLimit: 1
   issuerRef:
     name: trust-manager
--- a/bindata/trust-manager/resources/clusterrole_trust-manager-cluster-view.yml
+++ b/bindata/trust-manager/resources/clusterrole_trust-manager-cluster-view.yml
@@ -1,20 +1,18 @@
 ---
-kind: RoleBinding
+kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: trust-manager
-  namespace: cert-manager
   labels:
     app.kubernetes.io/name: cert-manager-trust-manager
     app.kubernetes.io/instance: cert-manager-trust-manager
     app.kubernetes.io/version: "v0.24.0"
     app.kubernetes.io/managed-by: cert-manager-operator
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
     app.kubernetes.io/part-of: cert-manager-operator
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: trust-manager
-subjects:
-  - kind: ServiceAccount
-    name: trust-manager
-    namespace: cert-manager
+  name: trust-manager-cluster-view
+rules:
+  - apiGroups:
+      - "trust.cert-manager.io"
+    resources:
+      - "bundles"
+    verbs: ["get", "list", "watch"]
--- a/pkg/controller/trustmanager/rbacs.go
+++ b/pkg/controller/trustmanager/rbacs.go
@@ -45,6 +45,11 @@ func (r *Reconciler) createOrApplyRBACResources(trustManager *v1alpha1.TrustMana
 		return err
 	}
 
+	if err := r.createOrApplyClusterViewClusterRole(trustManager, resourceLabels, resourceAnnotations); err != nil {
+		r.log.Error(err, "failed to reconcile cluster-view clusterrole resource")
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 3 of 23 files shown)


pkg/controller/trustmanager/certificates.go
```diff
@@ -113,7 +113,8 @@ func certificateModified(desired, existing *certmanagerv1.Certificate) bool {
 		!slices.Equal(desired.Spec.DNSNames, existing.Spec.DNSNames) ||
 		desired.Spec.SecretName != existing.Spec.SecretName ||
 		!ptr.Equal(desired.Spec.RevisionHistoryLimit, existing.Spec.RevisionHistoryLimit) ||
-		!reflect.DeepEqual(desired.Spec.IssuerRef, existing.Spec.IssuerRef) {
+		!reflect.DeepEqual(desired.Spec.IssuerRef, existing.Spec.IssuerRef) ||
+		!reflect.DeepEqual(desired.Spec.SecretTemplate, existing.Spec.SecretTemplate) {
```
bharath-b-rh: We should be checking privateKey too, which is also a new addition.

pkg/controller/trustmanager/rbacs.go
```diff
@@ -108,6 +113,44 @@ func appendSecretTargetRules(clusterRole *rbacv1.ClusterRole, secretTargets v1al
 	})
 }
 
+const clusterReaderAggregateLabel = "rbac.authorization.k8s.io/aggregate-to-cluster-reader"
```
bharath-b-rh, line 116: Static manifest already has this label added, what would be need for adding it here again?

pkg/controller/trustmanager/webhooks.go
```diff
@@ -63,12 +63,14 @@ func updateWebhookClientConfig(webhookConfig *admissionregistrationv1.Validating
...
-// cert-manager CA injection annotation. The CA injection annotation references
-// the Certificate resource by namespace/name and is constructed dynamically.
+// cert-manager CA injection annotation. From trust-manager v0.25.0 the webhook
+// injects CA from the TLS Secret (inject-ca-from-secret) rather than the
+// Certificate resource. The annotation is constructed dynamically.
 func updateWebhookAnnotations(webhookConfig *admissionregistrationv1.ValidatingWebhookConfiguration, resourceAnnotations map[string]string) {
 	mergedAnnotations := make(map[string]string, len(resourceAnnotations)+1)
 	maps.Copy(mergedAnnotations, resourceAnnotations)
-	mergedAnnotations["cert-manager.io/inject-ca-from"] = fmt.Sprintf("%s/%s", operandNamespace, trustManagerCertificateName)
+	delete(mergedAnnotations, "cert-manager.io/inject-ca-from")
```
bharath-b-rh: Do we really need this cleanup? The resource is built from static manifest from scratch, it wouldn't have the older annotation set, correct? And since we use update (during upgrades) the annotation on the existing resource would also get overwritten. So I really don't see the need to add this. PLMK if the understanding is incorrect.

bindata/trust-manager/resources/certificate_trust-manager.yml
```diff
@@ -7,14 +7,19 @@ metadata:
...
 spec:
   commonName: "trust-manager.cert-manager.svc"
   dnsNames:
     - "trust-manager.cert-manager.svc"
   secretName: trust-manager-tls
+  secretTemplate:
+    annotations:
+      cert-manager.io/allow-direct-injection: "true"
+  privateKey:
+    rotationPolicy: Always
```
bharath-b-rh, line 22: Upstream also added option to configure duration based on helm value `app.webhook.tls.certificate.duration`, which we should be also considering.

Also, upstream added another option to create `CertificateRequestPolicy` resource, if approver policy is enabled, which we should also be considering, since approver-policy will be TP in this release.

bindata/trust-manager/resources/clusterrole_trust-manager-cluster-view.yml
```diff
@@ -0,0 +1,18 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    app.kubernetes.io/name: cert-manager-trust-manager
+    app.kubernetes.io/instance: cert-manager-trust-manager
+    app.kubernetes.io/version: "v0.25.0"
+    app.kubernetes.io/managed-by: cert-manager-operator
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
```
bharath-b-rh, line 10: for this to work, we also need to create `deploy/charts/trust-manager/templates/clusterrole-aggregate.yaml`, which otherwise has value I think.

bharath-b-rh: Also the whole aggregation is an opt-in feature using the `aggregateClusterRoles` helm value, shouldn't we be also doing the same?

bharath-b-rh (commented):
Regarding `test/go.mod stays github.com/cert-manager/trust-manager v0.20.3`. It should be fine updating the k8s packages to `v0.36.4` and also the `openshift/api` package to `release-5.0` commit.

pkg/controller/trustmanager/rbacs.go
arun717: **UpdateResourceLabels** replaces the object's labels (SetLabels) rather than merging them, so the aggregate label from the static manifest is dropped when we apply getResourceLabels().

We set rbac.authorization.k8s.io/aggregate-to-cluster-reader=true again afterward so cluster-reader still aggregates Bundle get/list/watch. Without this, the applied ClusterRole would lose that label.

pkg/controller/trustmanager/webhooks.go
arun717: Removed the delete step.

bindata/trust-manager/resources/certificate_trust-manager.yml
arun717: Helm’s default for app.webhook.tls.certificate.duration is empty, so omitting duration matches upstream. Making it configurable would be a TrustManager API change; I would keep that out of this operand bump. Maybe I can include it in PR #494 where we have api changes. Similarly for CertificateRequestPolicy I can handle in a different PR if needed.

bindata/trust-manager/resources/clusterrole_trust-manager-cluster-view.yml
arun717: Looks like we do not need another copy of clusterrole-aggregate.yaml.
That Helm file is this ClusterRole:

- name: trust-manager-cluster-view
- rules: get/list/watch on trust.cert-manager.io/bundles
- label: rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"

The operator already has that as **clusterrole_trust-manager-cluster-view.yml** and applies it in createOrApplyClusterViewClusterRole. 
There is no extra aggregator object in the chart, just this labeled ClusterRole.

Aggregation is performed by OpenShift’s existing **cluster-reader**, which already selects that label. Creating a second YAML with the same content would be a duplicate.

Also, the aggregation part is configurable but  it is not opt-in in the usual sense.
Upstream Helm gates the ClusterRole with global.rbac.aggregateClusterRoles. That value defaults to true, so aggregation is on unless you turn it off (opt-out). 

The template only skips clusterrole-aggregate.yaml when you set it to false (or disable RBAC entirely with global.rbac.create: false).
Please do let me know what you think. As of now not making any code change related to this in the current PR.

arun717:
Manual live-cluster check against a 4.21 CI cluster (operator from this PR via `local-run`, operand image `quay.io/jetstack/trust-manager:v0.25.0`):

- TrustManager `cluster` Ready=True; deployment Available; container args have no `--filter-non-ca-certs`
- ValidatingWebhookConfiguration `trust-manager` has `cert-manager.io/inject-ca-from-secret=cert-manager/trust-manager-tls`
- ClusterRole `trust-manager-cluster-view` is read-only on `bundles` and labeled `rbac.authorization.k8s.io/aggregate-to-cluster-reader=true`; `cluster-reader` now includes those rules

Left unchecked: prow `e2e-operator-tech-preview` (lease 503; tests did not run). `e2e-operator` is green but that job excludes `TechPreview`.

